PR #4, opened 2026-04-10
Fix crash in GC

csm:
Values in the call stack aren't rooted, so they could be freed mistakenly, causing use-after-free crashes.


csm:
This also hides all aot_e2e tests under a feature flag. These tests take a very long time to run, we can do external validation, not run this on each CI run.

Issue #25, opened 2026-04-25
[bug] GC frees in-use objects


csm:
The GC regressed and now seems to be freeing in-use objects.

To reproduce, run the clojure-test-suite with a small heap:

```
cargo run --bin cljrs test --gc-soft-limit-mb 40 --src-path ./clojure-test-suite/test
```

Not including the backtrace here, since it's post-freed object access. The failure is: 

```
thread 'cljrs-main' (52851067) panicked at /Users/csm/src/clojurust/crates/cljrs-gc/src/lib.rs:299:13:
assertion `left == right` failed: GcPtr::get() on freed object! magic=0x221cf5e9e08cfe3d
  left: 2458109881480969789
 right: 14627333968688430831
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
cljrs: thread panicked: Any { .. }
```

csm:
Fixed with #26

PR #26, opened 2026-04-25
Simplify SetValue trace implementation using GcVisitor

csm:
## Summary
Refactored the `Trace` implementation for `SetValue` to use the `GcVisitor` trait directly instead of calling `trace()` on dereferenced values.

## Key Changes
- Added `use cljrs_gc::GcVisitor as _` import to bring the trait into scope
- Replaced `s.get().trace(visitor)` calls with `visitor.visit(s)` for both `Hash` and `Sorted` variants
- This change applies to both `SetValue::Hash` and `SetValue::Sorted` cases

## Implementation Details
The refactoring leverages the `GcVisitor` trait's `visit()` method, which likely handles the dereferencing and tracing internally. This is a cleaner API that reduces boilerplate and makes the code more consistent with the garbage collection framework's intended usage patterns.

https://claude.ai/code/session_01N1toG7Pu43nPT3HiVifq1U

### DIFF
--- a/crates/cljrs-value/src/value.rs
+++ b/crates/cljrs-value/src/value.rs
@@ -337,9 +337,10 @@ impl SetValue {
 
 impl cljrs_gc::Trace for SetValue {
     fn trace(&self, visitor: &mut cljrs_gc::MarkVisitor) {
+        use cljrs_gc::GcVisitor as _;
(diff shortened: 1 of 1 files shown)


csm:
Fixes #25

PR #40, opened 2026-05-08
fix(publish): eliminate duplicate version key in workspace Cargo.toml

csm:
The second sed pass prepended version = "=$VERSION" to workspace
dependency lines that already contained version = "0.1.0" at the end,
producing an invalid inline table with two version keys.

Fix: extend the single sed pass to cover Cargo.toml alongside
crates/*/Cargo.toml. One replacement, no duplicates. The separate
sed block that inserted the version prefix is removed entirely.

https://claude.ai/code/session_01Jh586cWmvxaw1JgHz3KyB4


csm:
I mean thanks for finally suggesting cargo-release instead of using sed. I guess I should have asked.

PR #46, opened 2026-05-09
Phase A async groundwork: tokio-backed futures, channels, and runtime

csm:
## Summary

This PR introduces Phase A async infrastructure for clojurust, gating all async-specific code behind an `async` Cargo feature. The main changes are:

1. **Refactored `CljxFuture`** to support dual backends:
   - Without `async` feature: original `Mutex<FutureState>` + `Condvar` (OS thread parking)
   - With `async` feature: `tokio::sync::Mutex` + `tokio::sync::Notify` for executor-friendly async, plus a shadow `Mutex<FutureState>` for blocking callers that don't need a tokio runtime handle

2. **New `CljxChannel` type** (async-only):
   - Wraps `tokio::sync::mpsc` for core.async-style channels
   - Capacity support (rendezvous mode stubbed for Phase E)
   - Added to `Value` enum behind `#[cfg(feature = "async")]`

3. **Tokio runtime integration**:
   - `cljrs` binary now wraps execution in `tokio::runtime::Builder::new_current_thread()` + `LocalSet` when `async` is enabled
   - Allows `tokio::task::spawn_local` to be called from within the interpreter without user setup

4. **API improvements**:
   - Extracted `CljxFuture` methods: `blocking_deref()`, `blocking_deref_timeout()`, `complete()`, `fail()`, `cancel()`, `snapshot()`, `is_done()`, `is_cancelled()`
   - Added async-only `await_value()` method for use in `^:async` contexts
   - Renamed `(await ...)` builtin to `(await-agent ...)` to reserve `await` for the Phase B async special form

5. **Documentation**:
   - Added comprehensive README sections to all affected crates documenting the `async` feature
   - Clarified Phase A scope and Phase E roadmap for channels

## Key Implementation Details

- **Dual-state design for async futures**: The `sync_state` + `sync_cond` shadow mirrors the tokio-side state so blocking callers never need a runtime handle. Resolution atomically updates both.
- **Contention fallback**: When the tokio mutex is contended during resolution, a minimal `futures_lite_block_on` helper (parking_lot-style spin) acquires it without deadlock.
- **Feature propagation**: The `async` feature is properly threaded through all workspace crates (`cljrs-value`, `cljrs-builtins`, `cljrs-interp`, `cljrs-stdlib`, `cljrs`) so it can be toggled at the top level.
- **Backward compatibility**: Without the feature, all code paths are identical to before; `Value::Channel` doesn't exist, and `CljxFuture` uses the original synchronous backend.

## Testing

- Existing `deref` and agent tests updated to use new `CljxFuture` API
- Tests continue to pass with and without the `async` feature

https://claude.ai/code/session_01VFna25M6bJjLRjvvDehGHu

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -265,6 +265,7 @@ dependencies = [
  "cljrs-value",
  "miette",
  "rustyline",
+ "tokio",
  "tracing",
  "tracing-subscriber",
 ]
@@ -285,6 +286,7 @@ dependencies = [
  "rand",
  "regex",
  "rpds",
+ "tokio",
  "uuid",
 ]
 
@@ -371,6 +373,7 @@ dependencies = [
  "cljrs-types",
  "cljrs-value",
  "regex",
+ "tokio",
  "uuid",
 ]
 
@@ -437,6 +440,7 @@ dependencies = [
  "cljrs-reader",
  "cljrs-types",
  "cljrs-value",
+ "futures-util",
  "lazy_static",
  "tokio",
 ]
@@ -464,6 +468,7 @@ dependencies = [
  "regex",
  "rpds",
  "thiserror",
+ "tokio",
  "uuid",
 ]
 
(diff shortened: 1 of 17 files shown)


csm:
This is better left as a library, similarity to clojure.core.async.

PR #88, opened 2026-05-25
Phase F: higher-level async utilities (join-all, mult, merge, reduce, pmap)

csm:
Add all Phase F clojure.core.async functionality:

Native Rust builtins (builtins.rs):
- join-all: await a seq of futures, return a vector of all results
- thread-call: run a thunk async, deliver result to a buffered channel
- onto-chan!: seed a channel from a collection, then close it
- to-chan!: same as onto-chan! but returns channel before seeding finishes
- mult: broadcast a source channel to registered tap channels
- tap!, untap!, untap-all!: manage mult tap lists

New CljMult native object (channel.rs):
- CljMult holds (tap_channel, close_on_done) pairs in a Mutex<Vec>
- Implements NativeObject and Trace so GC can traverse tap channels

Clojure-level (core_async.cljrs):
- async-pmap: map an ^:async fn over coll concurrently via join-all
- thread macro: sugar over thread-call
- merge: fan-in multiple channels into one, close output when all inputs drain
- reduce: fold all channel values with (f acc v), ^:async
- into: drain channel into collection via reduce, ^:async

Async loop support (eval_async.rs):
- eval_loop_async: properly evaluates loop/recur inside ^:async bodies so
  await yields correctly rather than falling through to the sync evaluator

Bug fix (env.rs):
- refer_named now uses insert (always overrides) instead of or_insert_with
  so (require :refer [name]) correctly shadows clojure.core defaults

https://claude.ai/code/session_01RC38eGXBufm9UQVB4NLgzC

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,7 +34,7 @@ jobs:
         run: cargo test --workspace
 
       - name: Clojure Tests
-        run: cargo run -p cljrs compile -o test-suite --test ./clojure-test-suite/test && CLJRS_GC_STATS=- ./test-suite
+        run: cargo run -p cljrs compile -o test-suite --test ./clojure-test-suite/test && CLJRS_GC_STATS=- CLJRS_GC_SOFT_LIMIT_MB=1024 ./test-suite
 
       - name: Graph sample (region-alloc smoke test)
         env:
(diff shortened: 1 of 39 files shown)


csm:
Oof. This PR and its failures really exposed some issues with the GC implementation. This will take a few revisions to get it right

PR #94, opened 2026-05-27
Add Conway's Game of Life as bump-allocator showcase sample

csm:
Each generation materialises ~8 tiny [row col] neighbour vectors per
live cell, checks them, and immediately discards them.  None escape
their call site so the escape analyser can region-allocate them; with
the default settings (~600 cells, 150 generations) roughly 650 000
such vectors are created and bulk-reclaimed by the bump allocator.

Also adds a "Game of Life (bump-allocator showcase)" CI step that AOT
compiles and runs samples/life.cljrs with CLJRS_GC_STATS=- so the
region vs heap allocation split is visible in every CI run.

https://claude.ai/code/session_01EASUtDqdSqepMWENU6uxc4

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,6 +41,11 @@ jobs:
           CLJRS_GC_STATS: "-"
         run: cargo run -p cljrs compile -o graph-sample samples/graph.cljrs && ./graph-sample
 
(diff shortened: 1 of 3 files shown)


csm:
Loving it when “let’s add a neat demo” turns into “oh let’s fix this fatal compiler bug.”

PR #105, opened 2026-05-29
fix: remove cljrs-stdlib build-dependencies that deadlock cargo publish

csm:
cargo's workspace publish ordering tracks each package's in-degree (how many workspace deps must be confirmed before it can publish). cljrs-stdlib listed cljrs-ir, cljrs-eval, cljrs-interp, cljrs-value, and cljrs-reader in BOTH [dependencies] and [build-dependencies]. Cargo increments the in-degree counter once per occurrence across all dependency sections but decrements it only once per confirmation, so these five packages were double-counted.  cljrs-stdlib's counter never reached zero even after all its dependencies were confirmed, which permanently blocked cljrs-stdlib and every crate that depends on it -- exactly the 17-package deadlock we saw on every publish run.

The [build-dependencies] existed solely to support the `prebuild-ir` feature, which pre-lowers clojure.core to IR at compile time and embeds the result via include_bytes!.  That feature was never enabled anywhere in the workspace (not in the cljrs binary, not in CI, not in any dependent crate), making it dead code that only existed to break publishing.

Remove the `prebuild-ir` feature and the entire [build-dependencies] section from cljrs-stdlib/Cargo.toml.  Trim the corresponding dead code from build.rs and lib.rs: the static IR bundles, both versions of load_prebuilt_compiler_ir, and the prebuilt-IR fast-path in standard_env().  The always-active fallback path (runtime compiler loading on a background thread) is unchanged.

https://claude.ai/code/session_01Segg9ZackfiKLqoJqfHjwb


csm:
Somewhat skeptical of the confidence Claude showed here. But this is genuinely dead code now (we tried having the IR lowerer be in Clojure, which didn’t work out).

PR #113, opened 2026-05-30
feat(net): implement networking-plan.md Phase A — cljrs-net crate + TCP client

csm:
Adds the `cljrs-net` crate (Phase A of networking-plan.md):

- `TcpStreamResource` implementing the `Resource` trait — Arc-backed,
  deterministic close via `AbortHandle`s for reader/writer tasks
- `connect` builtin: `(connect {:host h :port p})` returns a capacity-1
  promise channel that yields a connection map `{:in :out :remote-addr
  :local-addr :resource}` once the TCP handshake completes, or a
  `Value::Error` on failure
- `close` builtin: closes `:in`/`:out` channels and aborts both tasks
- Reader task: chunks socket data into `byte-array` values on `:in`,
  closes `:in` at EOF or error
- Writer task: drains `:out`, writes `byte-array`/string values to the
  socket, calls `shutdown(Write)` when `:out` closes (TCP half-close)
- `clojure.rust.net.tcp` namespace (native + Clojure source)
- `clojure.rust.net` umbrella namespace with transport-agnostic `connect`
  and `close` (defaults to `:transport :tcp`)

Also:
- Adds `pub async fn CljChannel::take()` to `cljrs-async` (public
  counterpart to `put` for use by sibling crates)
- Wires `cljrs_net::init` into the CLI under a default-on `net` feature
- Integration tests: echo-server round-trip + connection-failure error path

https://claude.ai/code/session_0191HKMtgw6d6GhrJY3kPXLS

### DIFF
--- a/crates/cljrs-async/src/channel.rs
+++ b/crates/cljrs-async/src/channel.rs
@@ -253,6 +253,22 @@ impl CljChannel {
         }
     }
 
+    /// Asynchronously take a value from the channel, cooperatively yielding to
+    /// the `LocalSet` executor until a value is available or the channel closes.
+    /// Returns `Value::Nil` when the channel is closed and drained.
+    ///
+    /// This is the async counterpart to [`Self::take_blocking`] and the
+    /// building block other native crates use to consume channel values.
+    /// Must run within a Tokio `LocalSet` context.
+    pub async fn take(&self) -> Value {
+        loop {
+            if let Some(v) = self.try_take() {
+                return v;
+            }
+            tokio::task::yield_now().await;
+        }
+    }
+
     /// Asynchronously put `v` onto the channel, cooperatively yielding to the
     /// `LocalSet` executor until the value is accepted — buffered (capacity >= 1)
     /// or handed off to a taker (rendezvous). Resolves `true` on success, or
--- a/crates/cljrs-async/src/channel.rs
+++ b/crates/cljrs-async/src/channel.rs
@@ -22,7 +22,7 @@ use std::sync::{Condvar, Mutex};
 use std::time::Duration;
 
 use cljrs_gc::{GcPtr, MarkVisitor, Trace};
-use cljrs_value::{NativeObject, NativeObjectBox, Value};
+use cljrs_value::{NativeObject, NativeObjectBox, Value, gc_native_object};
 
 /// The native type tag reported by [`NativeObject::type_tag`] for channels.
 pub(crate) const CHANNEL_TAG: &str = "Channel";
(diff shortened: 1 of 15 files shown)


crates/cljrs-net/src/tcp.rs
```diff
@@ -0,0 +1,357 @@
...
+
+/// Deliver exactly one value on a capacity-1 promise channel and close it.
+async fn chan_deliver(ch: &GcPtr<NativeObjectBox>, v: Value) {
+    let _ = chan_ref(ch.get()).put(v).await;
+    chan_ref(ch.get()).close();
+}
+
+/// Async take from a channel: yields until a value is available or the channel
+/// closes. Returns `Value::Nil` when the channel is closed and drained.
+async fn chan_take(ch: &GcPtr<NativeObjectBox>) -> Value {
```
csm: Why is this needed in the net crate? This isn’t provided by the async crate?

crates/cljrs-net/src/lib.rs
```diff
@@ -0,0 +1,65 @@
...
+
+    if !globals.is_loaded(NS) {
+        globals.get_or_create_ns(NS);
+        globals.refer_all(NS, "clojure.core");
+        load_source(globals, NS, NET_SOURCE);
+        globals.mark_loaded(NS);
+    }
+}
+
+fn load_source(globals: &Arc<cljrs_env::env::GlobalEnv>, ns: &str, source: &str) {
```
csm: Can’t this function live in a common place? Does it need to be copied here?

crates/cljrs-async/src/channel.rs
```diff
@@ -253,6 +253,22 @@ impl CljChannel {
...
     }
 
+    /// Asynchronously take a value from the channel, cooperatively yielding to
+    /// the `LocalSet` executor until a value is available or the channel closes.
+    /// Returns `Value::Nil` when the channel is closed and drained.
+    ///
+    /// This is the async counterpart to [`Self::take_blocking`] and the
+    /// building block other native crates use to consume channel values.
+    /// Must run within a Tokio `LocalSet` context.
+    pub async fn take(&self) -> Value {
```
csm, line 276: What happens if the try_take never returns a value? Does this have to busy-loop like this?

csm: You're right that it's a polling loop — `yield_now()` gives up the task's executor slot each iteration (so other tasks including producers can run), but it re-polls at executor speed rather than parking until woken. If a value never arrives and the channel is never closed, it will loop forever.

This is actually the same approach `put` already uses (see the `try_put_buffered` and `rv_offer` loops above it), so `take` is at least consistent with the existing design.

The proper fix is to add a `tokio::sync::Notify` (or similar) inside `CljChannel` so blocked takers park and are woken up exactly when a putter delivers a value — eliminating the polling entirely. That's a meaningful refactor of `CljChannel` itself (the `Condvar`-based blocking paths would be unified with async notification). Happy to do that as a follow-up or as part of this PR if you'd prefer it here.

---

PR #140, opened 2026-06-05
docs: add Memory Management section to CLAUDE.md

csm:
Documents the two allocation strategies (tracing GC vs bump allocator),
how the Region bump allocator works, the StaticArena for program-lifetime
values, the allocation context stack, and the return-expression protocol.
Explicitly notes that the bump allocator is only available in AOT mode
(no-gc feature).

https://claude.ai/code/session_01HwJxVzink9jAC2BMHsUBEN


csm:
Sorry, Sonnet, you didn’t understand the assignment.

PR #156, opened 2026-06-10
Disable Clojure interpreter test in workflow

csm:
Comment out Clojure interpreter test step in workflow.


csm:
Interpreter GC rooting fixes make this work now, closing this and keeping the test.

PR #267, opened 2026-07-05
Fix native/JIT collection ops silently ignoring metadata-carrying values

csm:
rt_abi.rs's fast-path collection builtins (rt_get, rt_count, rt_first, rt_rest, rt_assoc, rt_conj, rt_dissoc, rt_disj, rt_nth, rt_contains, rt_seq, rt_is_empty, rt_is_vector, rt_is_map, rt_transient) matched directly on the raw Value variant, unlike their cljrs-builtins tree-walker counterparts, which all call .unwrap_meta() first. A metadata-carrying collection (any `(with-meta coll m)` — notably a protocol-implementation map built via `(with-meta {...} impl)`, the idiom Replicant's mutation-log renderer uses for its renderer/DOM-node objects) is `Value::WithMeta(inner, meta)`, so it fell through every fast path's `_` arm once the calling function crossed the JIT threshold: get silently returned nil, count/contains?/empty? reported wrong answers, first/rest/nth/seq returned nil/empty, and assoc/conj/dissoc/disj returned the value completely unchanged instead of updating it. Interpreted code (tree-walk and Tier-1 IR) was unaffected except for two inline predicates in ir_interp.rs's own dispatch_known_fn (IsVector/IsMap) that had the identical gap independent of JIT.

This was the actual root cause of the "stale children after promotion" defect (examples/cljrs-wasm/probes/39_stale_children_after_promotion.cljrs in Replicant's cljrs-port branch): an existing DOM node's text/children read back as empty once the reader function crossed --jit-threshold, because the renderer object being read (a with-meta-wrapped map) fell through rt_get's fast path. Verified against Replicant's full test suite under default settings (JIT enabled, no --ir-threshold workaround): stable 231-232 passed / 13-14 failed / 1 error across repeated runs, matching the project's own documented clean baseline (previously only reachable with --ir-threshold disabled) — no crashes, no data corruption, only the pre-existing, already-catalogued semantic issues unrelated to this defect.

Claude-Session: https://claude.ai/code/session_01KsszvNRDiEC42BipoTjVVE

### DIFF
--- a/crates/cljrs-compiler/src/rt_abi.rs
+++ b/crates/cljrs-compiler/src/rt_abi.rs
@@ -66,6 +66,18 @@ fn box_val(v: Value) -> *const Value {
 /// Safety: the returned pointer is valid until the active region is popped
 /// (i.e. until `rt_region_end`).  The IR analysis guarantees collection values
 /// are consumed before that point.
+/// Reattach `meta` (captured from an input collection via `get_meta()`,
+/// before `unwrap_meta()` peeled it off to dispatch) to a freshly built
+/// result, matching the identity-preserving builtins (`assoc`/`conj`/
+/// `dissoc`/`disj`) in cljrs-builtins, which all do the same.
+#[inline]
+fn apply_meta(v: Value, meta: Option<Value>) -> Value {
+    match meta {
+        Some(m) => v.with_meta(m),
+        None => v,
+    }
+}
+
 #[inline]
 fn box_coll_val(v: Value) -> *const Value {
     if cljrs_gc::region::region_is_active() {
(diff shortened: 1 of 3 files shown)


crates/cljrs-compiler/src/rt_abi.rs
```diff
@@ -1201,7 +1201,7 @@ pub unsafe extern "C" fn rt_alloc_cons(head: *const Value, tail: *const Value) -
 /// Both pointers must be valid.
 #[unsafe(no_mangle)]
 pub unsafe extern "C" fn rt_get(coll: *const Value, key: *const Value) -> *const Value {
-    let coll = unsafe { val_ref(coll) };
+    let coll = unsafe { val_ref(coll) }.unwrap_meta();
```
csm, line 1216: Do we need to save and restore metadata on the retval for any of these?

csm: Good catch — yes, for `assoc`/`conj`/`dissoc`/`disj` (which cljrs-builtins already treats as identity-preserving: `builtin_assoc`/`builtin_conj`/`builtin_dissoc`/`builtin_disj` all capture `.get_meta()` before unwrapping and reattach it via `.with_meta()` to the result). Fixed in 38554e7: added an `apply_meta` helper and threaded meta capture/reattachment through `rt_assoc`, `rt_conj`, `rt_dissoc`, `rt_disj`, plus a regression test (`assoc_conj_dissoc_disj_preserve_metadata_under_jit`) that fails without it.

The read-only ops (`get`/`count`/`first`/`rest`/`nth`/`seq`/`contains?`/`empty?`/`vector?`/`map?`/`transient`) don't need this — their `builtin_*` counterparts never propagate the container's metadata onto what they return either (an extracted element or a seq view isn't the same collection), so `unwrap_meta()` alone already matches their behavior.

---

PR #284, opened 2026-07-18
Add cooperative gas metering across interpreter, IR interpreter, and JIT

csm:
### Motivation
- Provide a cooperative execution-credit budget to limit runaway/long-running evaluations across the tree-walker interpreter, Tier-1 IR interpreter, and generated native code.
- Surface a unified `GasExhausted` error when the budget is exhausted so callers can handle metering consistently.
- Integrate metering into the JIT code path so compiled code can check and early-return on exhaustion without violating semantics.

### Description
- Add a new `cljrs_env::gas` module with `GasMeter`, `GasGuard`, `charge(cost)`, and `take_exhausted()` to implement scoped, shareable metering.
- Introduce `EvalError::GasExhausted` to represent budget exhaustion at evaluation time.
- Instrument the tree-walking interpreter to charge one credit per form via `cljrs_interpreter::eval` and add `eval_with_gas(form, env, credits)` as a scoped entry that installs a `GasMeter`.
- Instrument the IR interpreter to charge a weighted checkpoint per basic block (`phis + insts + 1`) and return `EvalError::GasExhausted` on failure.
- Add JIT-side support: new runtime bridge `rt_gas_charge`, register it in the JIT builder and codegen, emit `emit_gas_checkpoint` in compiled functions (per IR basic block), and implement `rt_gas_charge` in `rt_abi.rs` to call the shared `gas::charge`.
- Make the JIT dispatch path check `cljrs_env::gas::take_exhausted()` after native calls and return `EvalError::GasExhausted` when set.
- Update READMEs and public exports to document and expose `eval_with_gas` and the gas module.
- Add unit/integration tests for the gas meter behavior.

### Testing
- Ran `cargo test -p cljrs-env` which executed the unit test `scoped_meter_charges_without_partial_consumption` and it succeeded.
- Ran `cargo test -p cljrs-interp` which executed the tests `ordinary_eval_is_unmetered`, `zero_credits_exhaust_before_evaluation`, and `recursive_work_exhausts_and_cannot_be_caught`, and they all succeeded.
- Ran the workspace test suite via `cargo test`; the new gas-related tests passed (no failures observed).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a59b8c252488332b5eafb26bd73d1a9)


crates/cljrs-eval/src/apply.rs
```diff
@@ -391,6 +391,10 @@ fn call_jit_native(
     // in ALLOC_ROOTS.  Clone it before the alloc frame drops.
     let result = unsafe { (*result_ptr).clone() };
 
+    if cljrs_env::gas::take_exhausted() {
+        return Err(cljrs_env::error::EvalError::GasExhausted);
+    }
```
csm, line 396: **Correctness (main issue): a nested dispatch consumes the `EXHAUSTED` flag, and the bridges then swallow the error — exhaustion can vanish entirely.**

For a nested native call (outer compiled fn → `rt_call`/`rt_call_ic` → `callback::invoke` → inner `call_jit_native`):

1. The inner `call_jit_native` hits this `take_exhausted()` and **consumes** the flag, returning `Err(GasExhausted)`.
2. `callback.rs:168-169` converts any non-`Thrown` `EvalError` — including `GasExhausted` — into `ValueError::Other(String)`.
3. `rt_call`'s `Err(_e) => rt_const_nil()` (rt_abi.rs:2652) degrades that to a plain nil.

At this point the flag is clear and, because charging is all-or-nothing, the meter still has credits. The outer function's remaining checkpoints can pass, so the outer call can **complete with `Ok` and a nil-infected result — no `GasExhausted` surfaced at all**, contradicting the PR's core guarantee.

Suggested fix: make the signal sticky — have dispatch *peek* rather than take, and clear the flag only in `GasGuard::install`/`Drop`. Then every subsequent dispatch boundary converts to `GasExhausted` no matter what the bridges did in between. (Optionally also zero the meter on first failure so cheaper blocks can't keep passing after exhaustion.)

Separate nit on ordering: this early return happens *before* `take_pending_exception()` below. If a throw-unwind and exhaustion coincide, the stashed thrown value is left behind and could misfire a later `rt_try` — exactly the hazard the comment at lines 398-401 exists to prevent. Drain both before returning.

---

crates/cljrs-interp/src/special.rs
```diff
@@ -1005,6 +1005,7 @@ fn eval_try(args: &[Form], env: &mut Env) -> EvalResult {
             result = Err(EvalError::Recur(args));
             None
         }
+        Err(EvalError::GasExhausted) => return Err(EvalError::GasExhausted),
```
csm, line 1008: **Catchability leaks once the error crosses a bridge boundary.**

This rethrow (and the Tier-1 `eval_ir_try_catch_finally` pass-through) correctly keeps `GasExhausted` uncatchable on the direct path — but once the error crosses a callback/bridge boundary it reappears as `ValueError::Other` / `EvalError::Runtime("gas exhausted")`, and non-`Thrown` errors are converted to values and matched against catch clauses just below (lines 1013-1015), so `(catch :default e …)` catches it. A budget-limited program that routes work through a higher-order builtin (e.g. a predicate passed to `filter` running under `rt_count_filter`-style bridges) or a natively-compiled `try` can catch the exhaustion and keep running on the remaining credits.

The `recursive_work_exhausts_and_cannot_be_caught` test only covers the pure tree-walker path where the invariant holds. Giving `ValueError` a dedicated exhaustion variant (or relying on a sticky `EXHAUSTED` flag, per the `apply.rs` comment) would close this.

Minor: this early `return` also skips the `finally` body (line 1037-1039), while the async `eval_try_async` version flows through and runs it. Effectively equivalent (any `finally` eval fails its first charge immediately), but the two should pick one shape.

---

crates/cljrs-interp/src/eval.rs
```diff
@@ -21,6 +21,9 @@ use regex::Regex;
 
 /// Evaluate a `Form` in the given `Env`.
 pub fn eval(form: &Form, env: &mut Env) -> EvalResult {
+    if !cljrs_env::gas::charge(1) {
+        return Err(EvalError::GasExhausted);
+    }
```
csm, line 26: **Performance: every form evaluation now pays a TLS access + `RefCell` borrow + `last()` even when unmetered** — which is the default, since nothing in the CLI/REPL/nREPL installs a meter yet. This is the hottest path in the interpreter.

A cheap fast path would help, e.g. a `Cell<bool>`/`Cell<usize>` "meter installed" flag (or `Cell<Option<NonNull<GasMeter>>>` for the innermost meter) so the common case is a single TLS load and compare. Would be good to see a quick benchmark (e.g. `samples/life.cljrs`) before/after either way — the PR has no performance numbers.

---

crates/cljrs-compiler/src/codegen.rs
```diff
@@ -548,6 +549,8 @@ impl<'a, 'b, M: Module> FunctionTranslator<'a, 'b, M> {
                 }
             }
 
+            self.emit_gas_checkpoint((ir_block.phis.len() + ir_block.insts.len() + 1) as u64)?;
```
csm, line 552: **Performance: every basic block of all compiled code now contains an out-of-line call to `rt_gas_charge` plus a conditional branch — permanently, even when no meter is installed** (the default). That's a standing tax on exactly the code the JIT exists to make fast, and it adds one extra block pair per IR block to every function.

Consider gating checkpoint emission on a compilation option (compile metered variants on demand when a budget is active), or at minimum benchmark the JIT-heavy samples before merging.

Also worth noting: `translate_poll_fn` gets no checkpoints, so compiled async code is unmetered at the native tier (adding one there would actually be wrong as-is, since poll functions return `i32`, not a pointer — the nil return in `emit_gas_checkpoint` only matches the regular `ptr` signature from `declare_function`). Fine as a limitation, but it deserves a line in the gas module docs.

---

crates/cljrs-compiler/src/codegen.rs
```diff
@@ -1371,6 +1374,22 @@ impl<'a, 'b, M: Module> FunctionTranslator<'a, 'b, M> {
         self.builder.ins().call(func_ref, &[]);
     }
 
+    /// Charge a weighted basic block and return nil immediately on exhaustion.
+    fn emit_gas_checkpoint(&mut self, cost: u64) -> CodegenResult<()> {
```
csm, line 1381: **Semantics note (bounded, worth documenting): after a callee exhausts, the caller's current basic block still runs to completion with a fabricated nil** standing in for the callee's real result. Side effects in that window execute, and a natively-compiled `try` treats the nil as a normal body result (no pending exception is stashed), so the catch is skipped and execution continues until the next failing checkpoint. This is inherent to the nil-return protocol and bounded to one block, but it should be called out as a known semantic of the cooperative design in the gas module docs.

The mechanics here are otherwise sound: `seal_all_blocks()` at the end of `translate` covers the new blocks, and regular compiled functions always return `ptr_type` (line 244), so `return_(&[nil])` is ABI-safe. Verified the `jit_specialization` end-to-end suite passes on this branch.

---

crates/cljrs-eval/README.md
```diff
@@ -63,6 +63,7 @@ src/
 pub use cljrs_interp::eval::eval;
 pub use cljrs_env::env::{Env, GlobalEnv};
 pub use cljrs_env::error::{EvalError, EvalResult};
+pub use cljrs_interp::eval::{eval, eval_with_gas};
```
csm, line 66: This snippet now lists `eval` twice — line 63 still has the old `pub use cljrs_interp::eval::eval;`. The old line should be replaced by this one (as `lib.rs` correctly does); as real code the duplicate import would be a compile error (E0252).

---

crates/cljrs-env/src/gas.rs
```diff
@@ -0,0 +1,91 @@
...
+thread_local! {
+    static ACTIVE: RefCell<Vec<Arc<GasMeter>>> = const { RefCell::new(Vec::new()) };
+    static EXHAUSTED: Cell<bool> = const { Cell::new(false) };
+}
+
+/// Installs a meter for the dynamic extent of an evaluation.
+pub struct GasGuard;
+
+impl GasGuard {
+    pub fn install(meter: Arc<GasMeter>) -> Self {
```
csm, line 44: A few things on the guard/meter design:

- **`GasGuard` should be `#[must_use]`** — `GasGuard::install(meter);` without a binding drops immediately and silently uninstalls the meter.
- **Nesting escapes the outer budget**: `charge` only hits the innermost meter (`.last()`), so a nested `eval_with_gas` runs free of the outer budget, and `install` clearing `EXHAUSTED` here can also wipe an outer guard's pending exhaustion signal before it's been observed. If nesting isn't meant to be supported, consider asserting/documenting that; if it is, charging every meter in the stack is probably the intended semantics.
- **Thread-locality**: `GasMeter` is `Arc`+atomic (cross-thread ready) but installation is thread-local, so work spawned onto other threads (futures, agents, `pmap`) escapes the budget entirely. Worth documenting as a known limitation in the module docs.

---

crates/cljrs-interp/tests/gas_meter.rs
```diff
@@ -0,0 +1,40 @@
...
+
+#[test]
+fn zero_credits_exhaust_before_evaluation() {
+    assert!(matches!(
+        cljrs_interp::eval::eval_with_gas(&form("1"), &mut env(), 0),
+        Err(EvalError::GasExhausted)
+    ));
+}
+
+#[test]
```
csm, line 33: All three tests exercise the tree-walker only. Given the point of the PR is cross-tier unification, coverage is missing for:

- exhaustion inside a Tier-1/JIT-executed function, asserting `GasExhausted` comes back to the caller (this would have caught the flag-consumption issue flagged in `apply.rs`);
- a nested native-call case (outer compiled fn calling an inner fn that exhausts);
- a higher-order-builtin case (e.g. a predicate that exhausts under `filter`), which currently degrades the error to a catchable `Runtime` string.

At least the first seems essential before merge.

---

csm (commented):
## Review: cooperative gas metering across tiers

**Overview.** The core plumbing is solid and well-integrated with the existing sentinel/deopt machinery: all-or-nothing charging via `fetch_update`/`checked_sub` is correct, block-cost weights match between Tier-1 and native code, `seal_all_blocks()` covers the new checkpoint blocks, and the nil-return is ABI-safe because regular compiled functions always return `ptr_type`. Verified on this branch: `cargo check --workspace` clean, all gas tests pass, and the `jit_specialization` end-to-end suite passes.

**Main issue (see inline comment on `apply.rs`):** the exhaustion signal doesn't survive bridge boundaries. `callback::invoke` degrades `GasExhausted` to `ValueError::Other`, the `rt_call` bridges swallow that into nil, and a *nested* `call_jit_native` consumes the `EXHAUSTED` flag — so for nested native calls, exhaustion can silently become an `Ok` result with a nil-infected value, defeating the PR's stated guarantee. The fix is cheap: peek instead of take, clear only in `GasGuard` install/drop.

**Also flagged inline:**
- catchability leak: the error reappears as a catchable `Runtime("gas exhausted")` after crossing a bridge, undermining the uncatchable invariant `eval_try` carefully maintains;
- performance: the unmetered path (the default) now pays per-form TLS+`RefCell` overhead in the tree-walker and a per-basic-block runtime call in *all* compiled code, permanently — no benchmarks in the PR;
- bounded fabricated-nil window after a callee exhausts (worth documenting);
- `cljrs-eval` README now lists `eval` twice;
- `GasGuard` `#[must_use]`, nested-meter and thread-locality semantics;
- test coverage is tree-walker-only; at least one cross-tier test seems essential.

**Verdict:** I'd hold merge for the flag-consumption fix, the README duplicate, and one cross-tier test. The catchability/perf/doc items are reasonable as documented limitations or follow-ups.

---

crates/cljrs-async/src/state_machine.rs
```diff
@@ -197,6 +200,7 @@ impl Future for CompiledAsyncTask {
             }
             POLL_READY => Poll::Ready(Ok(this.sm.pending.clone())),
             POLL_THREW => Poll::Ready(Err(EvalError::Thrown(this.sm.pending.clone()))),
+            POLL_GAS_EXHAUSTED => Poll::Ready(Err(EvalError::GasExhausted)),
```
csm, line 203: Two issues with the new poll-fn metering:

**1. No meter is ever installed around the poll, so this path is either dead or charges the wrong budget.** `CompiledAsyncTask::poll` installs an eval-context guard (line 187-189) for the detached executor thread, but nothing installs a `GasGuard` there. The gas `ACTIVE` stack is thread-local, so in the normal case the executor thread's stack is empty, every `rt_gas_charge` succeeds for free, and compiled async remains unmetered — the new checkpoints are pure per-block overhead. Conversely, if a *metered* evaluation's guard happens to be active on this thread when the LocalSet polls an unrelated task, that task silently drains the wrong budget and can be killed with `POLL_GAS_EXHAUSTED` on someone else's credits.

The natural fix mirrors the `eval_ctx` mechanism right above: carry an optional `Arc<GasMeter>` in `CljxStateMachine` (captured at spawn time from the spawning evaluation's active meter) and install a `GasGuard` around each poll.

**2. This arm doesn't consume the `EXHAUSTED` flag.** The failed charge inside the poll fn set the thread-local `EXHAUSTED` flag on the executor thread, and returning `Err(GasExhausted)` here leaves it set — a later unrelated `call_jit_native` on this thread will `take_exhausted()` and report a spurious `GasExhausted`. (This folds into the sticky-flag redesign suggested on `apply.rs`: if the flag is cleared only by `GasGuard` install/drop, this hazard goes away along with the nested-dispatch consumption bug.)

---

crates/cljrs-compiler/src/codegen.rs
```diff
@@ -1371,6 +1376,31 @@ impl<'a, 'b, M: Module> FunctionTranslator<'a, 'b, M> {
...
+        let cost = self.builder.ins().iconst(types::I64, cost as i64);
+        let charged = self.call_rt_1(self.rt.rt_gas_charge, cost)?;
+        let continue_block = self.builder.create_block();
+        let exhausted_block = self.builder.create_block();
+        self.builder
+            .ins()
+            .brif(charged, continue_block, &[], exhausted_block, &[]);
+        self.builder.switch_to_block(exhausted_block);
+        if self.is_poll_fn {
```
csm, line 1390: The poll-fn branch here correctly resolves the ABI concern from my earlier comment — poll functions return the `i32` `POLL_GAS_EXHAUSTED` code instead of the pointer nil, matching the state-machine ABI, and it reuses the existing `is_poll_fn` field and `POLL_*` constant pattern. Verified `cargo check --workspace` is clean and the `jit_specialization`, `jit_seam_correctness`, and `cljrs-async` suites pass on this head.

Note the checkpoint is currently inert for async in practice, though — see my comment on `state_machine.rs`: nothing installs a meter on the executor thread around polls.

---

csm (commented):
## Re-review of the update (`a9407db` / merge `8ea6332`)

The update adds gas checkpoints to compiled async poll functions with a new `POLL_GAS_EXHAUSTED` code — this correctly addresses the async-tier gap flagged earlier, including the ABI trap (poll fns return the `i32` code, not a pointer nil). Verified on the new head: `cargo check --workspace` clean; `jit_specialization`, `jit_seam_correctness`, and the full `cljrs-async` suite pass.

Two new concerns on this change (inline): no `GasGuard` is installed around polls on the executor thread, so the async checkpoints are either dead overhead or charge an unrelated evaluation's budget; and the `POLL_GAS_EXHAUSTED` arm leaves the `EXHAUSTED` flag set, which can surface as a spurious `GasExhausted` in a later dispatch on that thread.

**Still open from the first review:**
- the main flag-consumption / bridge-swallowing bug in `call_jit_native` (nested native-call exhaustion can silently become `Ok` with a nil result);
- catchability leak via `Runtime("gas exhausted")` after bridge crossings;
- duplicate `eval` line in `crates/cljrs-eval/README.md`;
- `#[must_use]` on `GasGuard`, nested-meter and thread-locality semantics;
- unmetered-path performance (per-form TLS in the tree-walker, per-block runtime call in all compiled code) with no benchmarks;
- cross-tier test coverage (still tree-walker-only).

The sticky-flag redesign (peek instead of take; clear only in `GasGuard` install/drop) would resolve the main bug and both new flag-lifetime hazards in one change.

---

PR #287, opened 2026-07-18
Introduce cooperative gas metering and exhaustion handling across eval, IR, JIT, and async state machines

csm:
### Motivation

- Add a cooperative execution-credit ("gas") mechanism so tree-walker, IR interpreter, and native/JIT tiers can be metered and preempt long-running or runaway computations.
- Surface a clear caller-facing error when budgets are exhausted (`EvalError::GasExhausted`) and ensure the signal propagates correctly across call boundaries, catches, and async compiled state machines.

### Description

- Introduce a new `cljrs_env::gas` module implementing `GasMeter`, `GasGuard`, thread-local active-meter stack, exhaustion flag, and helper APIs (`charge`, `is_exhausted`, `active_meter`, `take_exhausted`).
- Add `EvalError::GasExhausted` and wire `ValueError::GasExhausted` through serialization/deserialization so exhaustion can be represented as a catchable/transportable condition.
- Meter the tree-walking interpreter by charging one credit at form entry via `eval` and adding `eval_with_gas(form, env, credits)` to install a scoped meter.
- Meter the IR interpreter by charging a weighted checkpoint per basic block (`phis + insts + 1`) and returning `EvalError::GasExhausted` on failure.
- Add native/JIT support: expose `rt_gas_charge` runtime bridge, register the symbol for JIT/AOT, and emit gas checkpoints in the codegen (`emit_gas_checkpoint`) which branch-return a poll/early-sentinel on exhaustion.
- Propagate gas state into compiled async state machines by capturing the active meter on spawn (`spawn_state_machine`) and reinstalling it on polls via `GasGuard::install`; extend poll codes with `POLL_GAS_EXHAUSTED` and map it to `EvalError::GasExhausted` in `CompiledAsyncTask::poll`.
- Ensure native-tier exhaustion is detected after JIT dispatch (`apply::call_jit_native`) and surfaces as `EvalError::GasExhausted` before pending exceptions are handled.
- Update callback/bridge layers to map gas exhaustion into `ValueError::GasExhausted` and to preserve behavior in try/catch handling (`eval_try` / `eval_try_async`).
- Add README and API docs describing the `gas` module and the new `eval_with_gas` entry point.

### Testing

- Added unit tests in `crates/cljrs-env/src/gas.rs` for meter scoping and nested charging, and integration tests `crates/cljrs-interp/tests/gas_meter.rs` and `crates/cljrs-eval/tests/gas_meter_ir.rs` to validate interpreter and IR/JIT exhaustion semantics.
- Ran the test suite (`cargo test` across the workspace) including the new tests, and all tests passed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a59b8c252488332b5eafb26bd73d1a9)


csm:
## Review: cooperative gas metering

I read the full diff, traced the gas paths through all four tiers (tree-walker, IR interpreter, JIT, async), and ran the new tests (all pass).

### Verdict

The core mechanism is well built: a clean `GasMeter`/`GasGuard` design in `crates/cljrs-env/src/gas.rs` (atomic `checked_sub` so a failed charge never partially consumes, nested meters charged together, RAII scoping), the `GasExhausted` error plumbed through every error enum, serialization, nrepl, and CLI messages, and checkpoints at sensible granularities (per form, per IR block, per compiled block with matching weights). But "cooperative metering across tiers" is only partially achieved: exhaustion is imprecise and inconsistently catchable in the compiled tier, async work can escape the budget, and there's one likely soundness issue. Details, ranked:

**1. Compiled-tier gas returns can leak open bump regions (soundness).** `emit_gas_checkpoint` (`codegen.rs:1381`) emits a bare `return_` from *any* block, including blocks between `RegionStart` and `RegionEnd`. Every other early exit unwinds the thread-local region stacks: `rt_try` calls `unwind_regions_to` on the pending-exception paths (`rt_abi.rs:3749`, `:3765`). But a gas return stashes no pending exception — `rt_try` sees a plain nil and takes the success path, and `call_jit_native` (`apply.rs:394-403`) doesn't unwind either. A function exhausted mid-region leaves stale entries on `RT_REGION_STACK`; later allocations can land in a region that subsequently gets reset, which is exactly the dangling-value hazard the region machinery guards against everywhere else. The gas path needs the same unwind treatment as the exception path.

**2. Exhaustion in compiled code is imprecise — execution continues on a bogus nil.** A callee that fails a checkpoint returns nil, which looks like a legitimate value to its compiled caller. Because charges are block-weighted, the caller's next (cheaper) checkpoint can *succeed* — e.g. remaining=5, callee's cost-10 block fails, caller's cost-3 block passes — so the caller keeps executing real instructions, including side effects (`swap!`, IO), with nil substituted for the callee's result. The sticky `EXHAUSTED` flag guarantees the *final* result at `call_jit_native` is `GasExhausted`, so the caller never sees a wrong value — but side effects performed between exhaustion and the boundary run on corrupted data. Worth documenting at minimum; it's a real difference from the interpreter tiers, where exhaustion unwinds immediately.

**3. "Cannot be caught" is inconsistent across tiers.** The interp (`special.rs:1008`), async eval (`eval_async.rs:294`), and IR try/catch (which only catches `Thrown`) all correctly refuse to catch `GasExhausted`. Two holes:
- `rt_try` (`rt_abi.rs:3768-3771`) converts any non-`Thrown` `ValueError` — including `GasExhausted` from an interpreted try body — into `Value::Str("gas exhausted")` and **invokes the catch handler with it**. In the block-weighted case where the meter isn't fully drained, the catch body runs to completion with its side effects.
- Awaiting a future: `settle_future` (`eval_async.rs:63`) folds `Err(GasExhausted)` into a plain error value, so `(try @f (catch ...))` catches exhaustion as an ordinary throw, and the awaiting thread's `EXHAUSTED` flag (thread-local) is unset.

**4. Async and interpreted-concurrency escapes.** `spawn_state_machine` captures only `active_meter()` — the *innermost* meter — so under nested meters a compiled async task escapes the outer budget, contradicting the module doc ("inner evaluations cannot escape an outer budget"). Worse, the generic `spawn_future` path — interpreted `^:async` fns, `timeout`, `alts`, etc. in `cljrs-async/src/builtins.rs` — captures no meter at all: any interpreted async work spawned under a budget runs unmetered. If the goal includes containing runaway computations, spawning a future is currently a one-line jailbreak. The whole `ACTIVE` stack should be conveyed (like `dynamics::capture_current` does for bindings) on both paths.

**5. Stale sticky flag across sibling scopes.** `EXHAUSTED` is cleared only when the guard stack empties. If host code runs `eval_with_gas` (inner) inside an outer guard, an inner exhaustion sets the flag, the inner guard drops without clearing it, and if the host swallows the error and continues under the still-healthy outer meter, every subsequent `call_jit_native` spuriously returns `GasExhausted`. Clearing the flag when a non-outermost guard drops (or scoping the flag per-guard) would fix this.

**6. Native builtin work is unmetered.** Checkpoints exist only at form/block granularity. Long-running native paths — the eager seq bridges with Set/Map predicates (no `invoke`, zero charges for the whole iteration), native reduce over native fns, string building, `slurp` — consume no gas. Inherent to cooperative metering, but it's the largest practical loophole for the stated goal and deserves a documented limitation or charges inside the hot native iteration bridges.

**7. Always-on overhead.** Every compiled basic block now makes an unconditional call to `rt_gas_charge` plus a branch even when no meter is installed — real cost in hot JIT loops that previously had none — and the tree-walker pays a thread-local + `RefCell` borrow per form. Consider gating checkpoint emission on a metering mode, or an inline fast-path (load a cached thread-local "metering active" word before calling out).

**8. Mechanism without policy, and thin tests.** `eval_with_gas` is exported but nothing calls it — no CLI flag, no nrepl per-request budget, no Clojure-level API — so the feature isn't reachable from the product surface yet (fine for a first PR, but worth saying in the description). The tests only exercise tree-walker exhaustion: `tier1_function_exhausts_meter` exhausts its 100 credits long before background lowering could kick in (`force_eager_lowering` is exported but unused), so the IR block-weighted path, the JIT checkpoint/sticky-flag path, and async meter propagation are all effectively untested despite the description claiming IR/JIT validation.

Findings 1 and 4 are the ones I'd want addressed before merge; 2, 3, and 5 are semantic holes that are at least masked by the boundary check and could be documented-then-fixed; the rest are follow-up material.

csm:
## Re-review of b656dd8 ("Fix gas metering review issues")

I re-traced all eight findings against the fix commit and re-ran the test suites (`cljrs-env` gas units, `cljrs-interp`/`cljrs-eval`/`cljrs-async` gas integration tests, plus the full `cljrs-compiler`, `cljrs-eval`, and `cljrs-async` suites — all green). Verdict: **findings 1–5 are resolved**, 6–8 are properly documented as deferred limitations. Details:

**1. Region leak — resolved.** `rt_gas_charge` now calls `unwind_all_native_regions()` on a failed charge. The correctness argument is sound: with the new per-scope sticky flags, once any charge fails every subsequent checkpoint on the thread also fails, so *no* native frame will ever execute its matching `RegionEnd` — closing everything on `RT_REGION_STACK` is safe, including regions of outer compiled frames sitting below an interpreted bridge frame. One subtlety worth a code comment someday: this "a failed charge always sets a flag" guarantee relies on meters being monotonically decreasing (the post-failure re-scan for `remaining < cost` can never miss, because remaining only shrinks). It holds today; it would break silently if anyone adds a `refund`/`refill` API.

**2. Bogus-nil execution — resolved.** Two mechanisms together close the window: `charge()` early-returns false while any scope flag is set (so a cheaper later checkpoint can no longer sneak through), and cost-0 sticky polls are now emitted immediately after every `Call`/`CallKnown`/`CallDirect`/region-call, *before* the result is bound. A caller can no longer run side effects on a callee's exhaustion-nil.

**3. Catchability holes — resolved.** `rt_try` now returns nil without invoking the catch handler on a `GasExhausted` body error, and the new distinct `FutureState::GasExhausted` variant is re-raised as `EvalError::GasExhausted` at every await/deref surface (`await_value`, `eval_await`, `deref_value`, `builtin_deref`, `check_ready`/`rt_async_poll_ready` → `POLL_GAS_EXHAUSTED`, and the codegen `Await` branch that returns the poll code instead of routing through the thrown path). The non-catchability invariant is now consistent across all four tiers and across the await boundary.

**4. Async escapes — resolved.** `spawn_future` captures the *complete* meter stack (`active_meters()`) and reinstalls it scoped to each individual poll — correctly avoiding TLS leakage between sibling `LocalSet` tasks. Dropping the per-state-machine meter in favor of inheriting through `spawn_future` is a nice simplification: compiled and interpreted async now share one conveyance path.

**5. Stale sticky flag — resolved.** Per-guard flag stack, with the `inner_exhaustion_does_not_poison_healthy_outer_scope` regression test covering exactly the scenario I described.

**6–8. Documented deferrals — acceptable.** The `cljrs-interp` README now states the three limitations plainly: native builtins that don't re-enter the evaluator under-charge, compiled code pays the checkpoint call even when unmetered, and this is a host API with no CLI/nREPL policy yet. Two notes for the follow-up list:
- The always-on JIT overhead actually grew slightly in this commit (an extra cost-0 poll per call site, plus a flag scan at the top of every `charge`). Fine for correctness-first; the metering-mode fast path mentioned in the README is the right eventual fix.
- The IR test now forces eager lowering (good), and the async tests are solid, but there is still no test that drives a *JIT-compiled* function to exhaustion end-to-end — `emit_gas_checkpoint`'s generated code (including the new post-call polls and the region unwind path) is exercised only indirectly. Worth adding when the JIT tier is easy to force from a test.

Nothing new introduced by the fix that I could find: guard drop order in `install_meters` is order-insensitive (unit guards popping a shared stack), cost-0 charges are free on the unmetered path, and per-poll flag reinstallation is safe because an exhausted poll settles its task terminally.

Good work — this addresses the substance of every blocking finding. LGTM from my side.

PR #290, opened 2026-08-04
aot: discover harness binary via cargo JSON messages

ericjuta:
## Summary

AOT linking assumed the harness binary always lands at `target/release/cljrs-aot-harness`. That breaks when `CARGO_TARGET_DIR` is set or cargo places artifacts under a target-triple directory (e.g. `target/<triple>/release/...`).

This change:

- Runs `cargo build --release --message-format=json` for both `link_with_cargo` and `link_with_cargo_test_harness`
- Parses NDJSON compiler messages and selects the **last** `compiler-artifact` for bin `cljrs-aot-harness` with a non-null `executable`
- Uses that path for the copy to the final output (so layouts from `CARGO_TARGET_DIR` / triples are handled by cargo itself)
- Adds `serde_json` as a direct `cljrs-compiler` dependency for safe message parsing
- Adds unit tests covering last-match wins, non-bin/wrong-name/null/empty executable filtering, and empty stdout errors

Branched from tag `v0.1.228`.

## Test plan

- [x] `cargo test -p cljrs-compiler --lib` (85 passed)
- [ ] Confirm AOT link with default `target/` layout still finds the harness
- [ ] Confirm AOT link with `CARGO_TARGET_DIR` / triple layout finds the harness via JSON `executable`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -723,7 +723,7 @@ dependencies = [
 
 [[package]]
 name = "cljrs"
-version = "0.1.0"
+version = "0.1.228"
 dependencies = [
  "clap",
  "cljrs-async",
@@ -761,7 +761,7 @@ dependencies = [
 
 [[package]]
 name = "cljrs-async"
-version = "0.1.0"
+version = "0.1.228"
 dependencies = [
  "cljrs-builtins",
  "cljrs-env",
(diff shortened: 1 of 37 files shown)


ericjuta:
Rebased the AOT-only fix onto current `main` (without the v0.1.228 release version thrash) as https://github.com/csm/clojurust/pull/291. Please prefer that PR; this one can be closed.

ericjuta:
Superseded by main-based PR with AOT-only commit.

Issue #292, opened 2026-08-08
bug: inline defrecord protocol-method bodies cannot resolve the record's own fields


BuddhiLW:
## Summary

Inside an **inline** protocol method impl on `defrecord`, a bare reference to one of the record's own fields fails to resolve. `TODO.md:113` lists inline protocol impls as complete, and `(:field this)` / `extend-type` both work, so this looks like the field bindings simply are not installed as locals in the inline-method body scope.

## Repro

```clojure
(defprotocol P (f [this]))
(defrecord R [b] P (f [_] b))
(println (f (->R 7)))
```

```
$ cljrs run repro.cljc
Error:   × Unable to resolve symbol: b
```

Expected `7` (matches JVM Clojure).

## Matrix

Everything below on `c02b399e` (main), debug build.

| case | result |
|---|---|
| `(defrecord R [b] P (f [_] b))` - bare field | ❌ `Unable to resolve symbol: b` |
| `(defrecord R [b] P (f [_ x] (+ b x)))` - bare field + arg | ❌ same |
| `(defrecord R [b] P (f [this] b))` - `this` named rather than `_` | ❌ same |
| `(defrecord R [a b] P (f [_] a))` - two fields | ❌ `Unable to resolve symbol: a` |
| `(defrecord R [b] Object (toString [_] (str b)))` | ❌ same |
| `(defrecord R [b] P (f [this] (:b this)))` - keyword access | ✅ `7` |
| `(defrecord R [b])` + `(extend-type R P (f [this] (:b this)))` | ✅ `7` |
| `(:b (->R 7))` - field access from outside | ✅ `7` |

So the record itself and its fields are fine; only the **lexical scope of an inline method body** is missing them.

## Why it bites harder than it looks

The failure is not always a visible error. In a multi-file program the enclosing namespace's preamble is evaluated at startup, and a failure there appears to be swallowed - the program keeps running against a half-initialized namespace. In an AOT binary (`cljrs compile`) I got a silently **wrong** answer rather than a crash: the AOT binary printed `nil`, `cljrs run` on the same source printed `0`, and the correct answer was `183850`. The dispatch had quietly fallen through to a `:default` method.

That second part may deserve its own issue - happy to split it out if you'd prefer. Minimal shape:

```clojure
;; helper.cljc
(ns helper)
(def ^:dynamic *tuning* 31)
(defprotocol IWeigh (weigh [this x]))
[...]

PR #293, opened 2026-08-08
feat(aot): --require-fully-compiled gate on embedded source

BuddhiLW:
## The problem

`cljrs compile` compiles plain `defn` bodies to machine code. Everything `needs_interpreter` reports (`ns`, `require`, `defmacro`, `defonce`, `defprotocol`, `defrecord`, `defmulti`, `defmethod`, `extend-type`, `extend-protocol`) is written into the harness and pulled in with `include_str!`, **method bodies included**, even when the enclosing namespace compiles successfully. Namespaces that fail lowering or codegen fall back to source the same way, and pinned versioned deps are bundled as source unconditionally.

Today the only signal is an `eprintln!` during the build. On a two-file program whose `helper` namespace the compiler reported as successfully compiled (`[aot] compiled namespace helper -> src/ns_0_preamble.cljrs + __cljrs_ns_init_0`), plain `strings` on the output binary recovers:

```
(ns helper)
(defprotocol IWeigh (weigh [this x]))
(defrecord Weigher [bias]
  IWeigh
  (weigh [_ x] (+ (* x *tuning*) bias)))
(defmulti route (fn [m] (:kind m)))
(defmethod route :alpha [m] (weigh (->Weigher 7) (:v m)))
(defmethod route :default [_] 0)
```

That is fine as a default, since interpretation has to come from somewhere. It is less fine that there is currently no way to ask whether it happened.

## What this adds

An audit of the source-carrying channels, and a policy for what to do about it. Four steps, each usable on its own:

| step | function | purity |
|---|---|---|
| collect | `embedded_fragments`, the channels as raw `(channel, ns, text)` | pure |
| promote | `audit_source`, fragments to a `SourceAudit`, empties dropped | pure |
| decide | `SourceAudit::verdict(OpacityPolicy)` | pure |
| apply | `compile_file` | effectful |

```rust
pub fn audit_source(interpreted_source: &str, compiled_namespaces: &[CompiledNamespace],
                    versioned_bundled: &[(Arc<str>, String)]) -> SourceAudit;

impl SourceAudit {
    pub fn leaks(&self) -> &[SourceLeak];
    pub fn is_clean(&self) -> bool;
    pub fn total_bytes(&self) -> usize;
    pub fn verdict(&self, policy: OpacityPolicy) -> OpacityVerdict;
}

pub enum OpacityPolicy  { Report /* default */, RequireFullyCompiled }
pub enum OpacityVerdict { Clean, Tolerated, Rejected }
```

`compile_file` takes the policy rather than deciding for itself, and applies the verdict before the harness is written, so nothing lands on disk on rejection. A new source-carrying channel is one entry in `embedded_fragments`; a new policy is one `OpacityPolicy` variant.

## Behaviour

`OpacityPolicy::Report` is the default and preserves current behaviour, plus one summary line:

```
[aot] 2 source fragment(s) (162 bytes) embedded as readable text (--require-fully-compiled makes this an error)
```

`cljrs compile --require-fully-compiled` fails instead, listing what would have shipped and where:

```
Error:   × --require-fully-compiled: 2 source fragment(s) (162 bytes) would be
  │ embedded in the binary as readable Clojure text:
  │   • entry preamble: 35 bytes
  │   • namespace preamble (helper): 127 bytes
  │ Only plain `defn` bodies are compiled to machine code. Move the forms
  │ above out of the compiled unit, or drop --require-fully-compiled.
```

## Tests

- `tests/source_leak_audit.rs`: 13 property tests (proptest, added as a dev-dependency) covering the audit, its rendering, and the policy layer. The load-bearing one is soundness, that the audit is silent exactly when no readable text would ship, since a false negative there is a binary that leaks source while the gate reports success. Also byte accounting, per-channel cardinality, namespace attribution, monotonicity, and that only the strict policy can ever reject.
- `tests/aot_e2e.rs`: `require_fully_compiled_rejects_embedded_source` (a `defprotocol`/`defrecord` program) and `require_fully_compiled_accepts_plain_defn`.
- Checked with `cargo-mutants` over the new code, all mutants caught. The first pass left both `Display` impls unkilled. The report text is the only thing telling a caller which namespace to fix, so it is contract, and it is now covered.

`cargo fmt` and `cargo clippy --tests` are clean. Crate README updated in the same commit, per CLAUDE.md.

## Notes

`SourceLeak` and `OpacityPolicy` is how I named them. But, if you prefer otherwise, it's reasonable to also change them. The CLI flag and the library API can also land separately if you would rather take them one at a time.

Found #292 while measuring this.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -380,6 +380,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "bit-set"
+version = "0.8.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "08807e080ed7f9d5433fa9b275196cfc35414f66a0c79d864dc51a0d825231a3"
+dependencies = [
+ "bit-vec",
+]
+
+[[package]]
+name = "bit-vec"
+version = "0.8.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5e764a1d40d510daf35e07be9eb06e75770908c27d411ee6c92109c9840eaaf7"
+
+[[package]]
 name = "bitfields"
 version = "1.0.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -858,6 +873,7 @@ dependencies = [
  "cranelift-module",
  "cranelift-native",
  "cranelift-object",
+ "proptest",
  "target-lexicon",
  "tempfile",
  "wasm-encoder",
@@ -2452,6 +2468,18 @@ dependencies = [
 
 [[package]]
 name = "getrandom"
+version = "0.3.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "899def5c37c4fd7b2664648c28120ecec138e4d395b459e5ca34f9cce2dd77fd"
+dependencies = [
+ "cfg-if",
+ "libc",
+ "r-efi 5.3.0",
+ "wasip2",
+]
+
+[[package]]
+name = "getrandom"
 version = "0.4.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "300e883d756b2e4ec94e02791f39b04b522276138852cfc41d9fb7e904106099"
@@ -2459,7 +2487,7 @@ dependencies = [
  "cfg-if",
  "js-sys",
  "libc",
- "r-efi",
+ "r-efi 6.0.0",
  "rand_core 0.10.1",
  "wasm-bindgen",
 ]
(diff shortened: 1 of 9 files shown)


csm:
Biggest issue is that the —require-fully-compiled flag is ignored under —test and —wasm. Should error out if given those combinations instead of silently ignoring the flag. Also would be good to enforce the flag under those combinations, but at least should reject the unsupported combo for now.

BuddhiLW:
Good catch, and it was worse than ignored: the build reported success under a guarantee nothing had checked.

Pushed `ca0363f` — `resolve_opacity_policy` now decides the policy for a `compile` invocation and errors on both combinations before any work starts:

```
$ cljrs compile --test --require-fully-compiled . -o out
Error:   × --require-fully-compiled is not supported with --test: the test harness is
  │ not audited for embedded source

$ cljrs compile --target wasm --require-fully-compiled app.cljrs -o out.wasm
Error:   × --require-fully-compiled is not supported with --target wasm: the wasm
  │ backend is not audited for embedded source
```

4 unit tests over the combination matrix (including that the native path is not caught by the check) plus 3 CLI-level tests in `crates/cljrs/tests/compile_flag_combinations.rs` driving the built binary.

On enforcing it under those combinations, the two cases turned out not to be alike:

**`--test` cannot satisfy the flag, by design.** `compile_test_harness` pushes every test namespace to `interpreted_bundled` unconditionally (aot.rs:2573) — deftests can't be lowered. So a strict-policy test harness would always fail. Rejection looks like the terminal answer there, not a placeholder; happy to word the error that way if you agree.

**`--target wasm` is a different guarantee, not a missing one.** `compile_file_to_wasm` embeds no source at all, so there is nothing for `audit_source` to find. What it *does* do silently is drop code: a namespace the backend can't lower is skipped (aot.rs:470) and entry forms that `needs_interpreter` are filtered out of `__cljrs_main` (aot.rs:485), both left for an IR-interpreter tier that isn't wired up yet. So the wasm analogue of the flag is completeness — "fail if the module would omit any form or namespace" — rather than opacity.

That is a real gate and a small one, but it is a semantics call on your side, so I left it out of this PR rather than guess. If you want it, say the word and I will add it here (or as a follow-up), reusing the same collect/promote/decide/apply shape with a `BundleAudit` alongside `SourceAudit`.

PR #294, opened 2026-08-09
fix: expand #?@ reader-conditional splices in interpreted eval

BuddhiLW:
## Problem

Splicing reader conditionals (`#?@`) were silently dropped by the tree-walking and async interpreters:

```clojure
[1 #?@(:rust [:a :b]) 2]      ;=> [1 [:a :b] 2]   ; expected [1 :a :b 2]
{:a 1 #?@(:rust [:b 2]) :c 3}  ;=> read error      ; expected {:a 1 :b 2 :c 3}
```

Only the AOT path expanded them (`expand_reader_conds_deep`). In the interpreter, a lone splicing conditional reached per-element eval and collapsed to a single value; maps additionally failed the reader's parity check before expansion.

## Fix

Collection-literal and call-argument evaluation now resolve reader conditionals through a shared primitive (`expand_reader_conds_cow`) before evaluating elements — covering **vector, set, map, and call** positions in both the sync (`cljrs-interp`) and async (`cljrs-async`) evaluators. Map key/value parity is checked *after* expansion; the reader defers that check when a splicing conditional is present, since the element count is then platform-dependent.

- `cljrs-builtins`: add `expand_reader_conds_cow`; `select_reader_cond` pairs clauses via `chunks_exact(2)`
- `cljrs-interp` / `cljrs-async`: expand at every collection and call-arg site
- `cljrs-reader`: defer map parity to the evaluator when `#?@` is present (genuine odd-arity maps without a splice still error at read time)

## Behaviour after

| input | result |
|---|---|
| `[1 #?@(:rust [:a :b]) 2]` | `[1 :a :b 2]` |
| `#{1 #?@(:rust [2 3]) 4}` | `#{1 2 3 4}` |
| `{:a 1 #?@(:rust [:b 2]) :c 3}` | `{:a 1 :b 2 :c 3}` |
| `(vector 1 #?@(:rust [2 3]) 4)` | `[1 2 3 4]` |
| `[1 #?@(:clj [:a :b]) 2]` (no match) | `[1 2]` |

## Testing

- **Property-based** (`proptest`): a model-based oracle reimplementing the `:rust` selection rule, checked against `expand_reader_conds` over arbitrary interleavings of plain / non-splicing / splicing conditionals with mixed keys; plus length-law, no-`ReaderCond`-survives, idempotence, and identity invariants.
- **Eval-level cross-container** property: splicing evaluates identically to inlining across vector / set / data-list / call args.
- **Mutation testing** (`cargo-mutants`): reader-conditional functions and the parser predicate — all viable mutants caught (`select_reader_cond` refactored to `chunks_exact(2)` to remove equivalent index-arithmetic mutants).
- `cargo clippy --workspace -- -D warnings` clean.


csm:
# 1. Blocking: quoted maps with #?@ now silently corrupt data
form_to_value's Map arm never expands reader conditionals — it chunks the raw forms, and a splicing conditional converts to Value::Nil. Before this PR the reader rejected odd-form maps outright, so this was unreachable; the new deferral makes it parse. I ran it:

```clojure
'{:a 1 #?@(:rust [:b 2]) :c 3}
;; before PR: read error (loud)
;; after PR:  {:a 1, nil :c}   ← splice became a nil key, `3` silently dropped
```

Fix: run expand_reader_conds (or the cow) in form_to_value's Map arm before chunks(2).

# 2. Same area: form_to_value::Set and syntax_quote don't expand either
Pre-existing, but the same inconsistency (verified):
￼
```clojure
'#{1 #?@(:rust [2 3]) 4}   ;; => #{1 nil 4}
`(1 #?@(:rust [2 3]) 4)    ;; => (1 nil 4)   — sq_form has no ReaderCond arm; falls through to form_to_value → nil
```

Lists and vectors are handled in form_to_value, so maps/sets/syntax-quote are the odd ones out. Note the AOT path (expand_reader_conds_deep in aot.rs) recurses into Quote/SyntaxQuote and expands all four container types — so interp and AOT currently diverge on exactly these positions. For syntax-quote, a ReaderCond arm in sq_seq producing Segment::Many would fit the existing structure.

# 3. In-scope parity gap: async let*/loop* binding vectors
The PR claims both evaluators are covered, and sync let* does expand its binding vector (special.rs:548) — but eval_let_async/eval_loop_async use the raw vector. Verified:

```clojure
(let* [#?@(:rust [x 1])] x)
;; sync fn:  => 1
;; ^:async fn: => Runtime("let* binding vector must have even length")
```

(Sync loop* also lacks it, so binding-vector coverage is partial in 3 of 4 sites.) One shared expand-then-check-parity helper for binding vectors would close all of them.

PR #295, opened 2026-08-10
fix(interp): accept a metadata-carrying symbol in defonce

BuddhiLW:
`(defonce ^:private registry (atom {}))` fails with

```
defonce requires a symbol at position 0
```

while the identical `def` form works. `eval_defonce` extracted the name with a plain `require_sym`, which only matches a bare `Form::Symbol`; a metadata-carrying symbol reads as a `^` form and never matches. `eval_def` already solves this with the metadata-aware `extract_def_name`.

### Repro

```clojure
(defonce ^:private registry (atom {}))   ;; => error
(def     ^:private registry2 (atom {}))  ;; => ok
```

### Fix

`eval_defonce` now shares `def`'s `extract_def_name`, plus an explicit empty-args guard so the arity error stays distinct from the type error.

Metadata now reaches the var, the once-only binding semantics are untouched, and a non-symbol name is still rejected.

### Tests

New `crates/cljrs-interp/tests/defonce_metadata.rs`, 5 cases:

- `^:private` symbol is accepted and the var carries the metadata
- map-form metadata (`^{:doc "..."}`) is accepted
- re-evaluating a `defonce` with metadata does not rebind
- a bare symbol still works (no regression)
- a non-symbol name is still an error

Full `cljrs-interp` suite is green.

Found while porting a `.cljc` library to run on cljrs — the interpreter is what the AOT path routes `defonce` through, so both paths are covered by this.


BuddhiLW:
Adding property coverage to this branch turned up a bigger instance of the same bug, so the PR now carries a fix for it. Flagging the scope change rather than slipping it in.

The property: `def` and `defonce` must extract the same name from the same form, generated over arbitrary names and metadata shapes. Every single-mark example passed. The generator reached a stacked mark on its tenth case:

```clojure
(def ^:private ^{:doc "d"} x 1)   ; => def name must be a symbol
(def ^:private x 1)               ; fine
```

`extract_def_name` peeled one `Meta` layer, so a second mark left the inner form wrapped. It now recurses, merging each layer through the existing `merge_meta` with the outer mark winning, which is the reader's own outward-in order. `def`, `defn`, `defmacro` and `defonce` share that extraction, so the stacked spelling was broken in all four and is fixed in all four.

So the original one-line change was correct, but the bug was bigger than the report. `^:private ^{:doc "..."}` is common enough that others have probably hit it.

Added 4 properties and 3 examples (stacked form on both `def` and `defonce`, outer-wins on conflict). `cargo-mutants` over `extract_def_name`: 15 mutants, 2 caught, 13 unviable, none survived. Full `cljrs-interp` suite green. Branch is current with main.

If you would rather keep this PR to the reported line, say so and I will split the stacked-metadata fix into its own.

PR #298, opened 2026-08-10
feat(reader): read namespaced map literals (#:ns{…}, #::{…}, #::alias{…})

### LINKED ISSUE #296: reader: namespaced-map literals (#:ns{...}, #::{...}) are a read error
The reader has no case for `#:` , so every namespaced-map literal is a read error.

### Repro

```clojure
;; a.clj
(println (pr-str #:adt{:a 1 :b 2}))
```

```
$ cljrs run a.clj
Error: cljrs::read

  × read error: unknown # dispatch character: Some(':')
   ╭─[a.clj:1:18]
 1 │ (println (pr-str #:adt{:a 1 :b 2}))
   ·                  ┬
   ·                  ╰── here
   ╰────
```

The auto-resolved form fails identically:

```clojure
(println (pr-str #::{:a 1}))   ;; same read error
```

The fully-qualified spelling is fine, so this is purely the reader shorthand:

```clojure
(println (pr-str {:adt/a 1 :adt/b 2}))   ;; => {:adt/a 1, :adt/b 2}
```

### Where

`crates/cljrs-reader/src/lexer.rs` — the `#` dispatch match (around L200-272) handles `(`, `{`, `'`, `_`, `"`, `?`, `?@`, `#`, and tagged literals, then falls through to the error at L270. There is no `Some(':')` arm, and `grep -i namespaced crates/cljrs-reader/src` finds nothing, so the feature is absent rather than broken.

### Why it bites

Namespaced maps are what `pr-str` emits on the JVM whenever `*print-namespace-maps*` is true — which is the default in a `clojure.main` REPL. So any data file, EDN fixture, or golden-value file produced by a JVM process is unreadable by cljrs unless the producer remembers to bind `*print-namespace-maps*` to false. That is a silent trap: the producing side looks correct and the failure only appears on the cljrs side, in a file nobody edited by hand.

cljrs's *printer* is already consistent (it emits the fully-qualified form), so only the read direction is missing.

### Semantics to match

Per the JVM reader:

- `#:ns{:a 1, :b 2}` → `{:ns/a 1, :ns/b 2}`
- `#::{:a 1}` → keys qualified with the *current* namespace
- `#::alias{:a 1}` → keys qualified with the namespace `alias` resolves to
- Already-qualified keys inside are left alone; `:_/k` un-qualifies to `:k`
- Symbol keys follow the same rule as keyword keys

Found while running a `.cljc` test corpus across JVM / ClojureWasm / cljrs — the JVM-generated golden file was the thing cljrs could not read.

BuddhiLW:
Fixes #296.

`#:ns{…}`, `#::{…}` and `#::alias{…}` were a read error — the `#` dispatch match in `lexer.rs` had no `:` arm, so every namespaced map fell through to `unknown # dispatch character: Some(':')`. The fully-qualified spelling read fine and the printer already emits it, so the gap was read-only.

It bites hardest on data that crosses runtimes: `pr-str` on the JVM emits `#:ns{…}` whenever `*print-namespace-maps*` is true, which is the default in a `clojure.main` REPL. Any JVM-produced EDN or fixture was therefore unreadable here while the producing side looked entirely correct.

### Design

- **Lexer** emits `Token::NamespacedMap { ns, auto }` for the prefix and leaves the `{` in the stream, so the parser reads the body through the ordinary map path. `auto` marks the `#::` spellings; `ns` is the literal namespace, the alias, or `""` for bare `#::`.
- **Parser** parses the body, then hands it to a new pure `namespaced_map::qualify_keys`.
- **`namespaced_map.rs`** is the whole feature as one function over `Form` with no reader state.

Auto-resolved spellings lower to `FormKind::AutoKeyword` rather than being resolved in the reader. `::foo` already works that way, and `cljrs-interp`'s `resolve_auto_keyword` already handles both the bare and aliased shapes — so `*ns*` and its aliases stay the evaluator's business and namespace resolution keeps living in exactly one place. Nothing downstream of the reader changed.

### Semantics

Matching the JVM reader:

| Key in source | `#:adt{…}` | `#::{…}` | `#::al{…}` |
|---|---|---|---|
| `:a` | `:adt/a` | current-ns qualified | alias-qualified |
| `:other/a` | unchanged | unchanged | unchanged |
| `:_/a` | `:a` | `:a` | `:a` |
| `a` (symbol) | `adt/a` | read error | read error |
| `1`, `"s"`, … | unchanged | unchanged | unchanged |

Values are never touched, and the literal nests.

One deliberate divergence: an **auto-resolved map with a symbol key** is a read error. There is no auto-resolved symbol form to lower to, and leaving the key bare would read as success while producing the wrong map — the error names the workaround (spell the namespace). `#:ns{a 1}` with a literal namespace works normally.

`#:{…}` (no namespace) and a prefix not followed by a map are read errors rather than silent misparses.

### Tests

17 new: 10 on the pure qualifier (including "values are never touched" and the `/` division-symbol edge), 2 on the lexer prefix + its two error cases, 5 end-to-end through the parser. Workspace suite green at 1018.

Also verified against real data: a golden file generated by a JVM process with `*print-namespace-maps*` at its default — 49 entries, two of them `#:adt{…}` — now loads into cljrs, which is the exact trap the issue describes.

`crates/cljrs-reader/README.md` updated per the crate-README convention (file layout, the `Token` table, and a `namespaced_map` section carrying the table above).


csm:
Thanks for the thorough tests and documentation. The basic #:ns{...} path looks good, but I think this needs changes before it fully matches the JVM reader / fixes #296.

# 1. Validate the namespace prefix as an unqualified symbol

lexer.rs:525
read_symbol_chars() currently accepts invalid prefixes and silently produces malformed keys:

```clojure
#:foo/bar{:a 1} ;; => :foo/bar/a
#:1{:a 1}       ;; => :1/a
#:nil{:a 1}     ;; => :nil/a
```

The JVM rejects these because the prefix must be an unqualified Symbol.

# 2. Auto-resolved maps must support symbol keys
namespaced_map.rs:43-55
#296 says symbol keys follow the same rule as keyword keys, and the JVM accepts:

```clojure
#::{a 1}      ;; => {user/a 1}
#::alias{a 1} ;; => {resolved.alias/a 1}
```

This PR currently makes both read errors. If support is intentionally deferred, the PR should not claim to fully fix #296.

# 3. Deferred AutoKeyword keys break quote/IR paths
parser.rs:236, cljrs-builtins/src/form.rs:188-190, cljrs-ir/src/lower/anf.rs:252-256

```clojure
'#::{:a 1}
```

is converted through form_to_value into simple :a, not :user/a. In IR lowering, #::alias{...} becomes current-ns/alias/key instead of resolving the alias. The result currently depends on whether the form is quoted, interpreted, or lowered.

# Compatibility gaps

# 4. Allow whitespace between the prefix and {

lexer.rs:531-534
The JVM accepts:

```clojure
#:foo {:a 1}
#:foo, {:a 1}
#:: {:a 1}
#::alias {:a 1}
```

This implementation requires { immediately after the prefix. Whitespace should be skipped, but not comments.

# 5. / keys should be qualified

namespaced_map.rs:73-75
The JVM qualifies bare / keys:

```clojure
#:foo{/ 1}  ;; symbol key foo//
#:foo{:/ 1} ;; keyword key :foo//
```

/ is special during evaluation as division, but it is still an unqualified symbol during reader qualification. The current test pins the wrong behavior.

Minor
 • Cargo.toml adds proptest, but Cargo.lock is not updated.

BuddhiLW:
All five are fixed, and the third one turned out to be the tip of something bigger — flagging that rather than folding it in silently.

Every behaviour below was checked against JVM Clojure with the same input; where we differed, the JVM is what the tests now pin.

### 1. Prefix validation

The prefix is now a value object rather than a `(String, bool)` pair:

```rust
pub enum MapNs { Literal(String), CurrentNs, Alias(String) }
impl MapNs { pub fn parse(text: &str, auto: bool) -> Result<Self, String> }
```

`parse` reads it as an unqualified symbol, so `#:foo/bar{…}`, `#:1{…}`, `#:nil{…}`, `#:true{…}` are read errors. The payoff beyond the check itself: `(ns = "", auto = false)` — the bare `#:{…}` with no namespace to qualify with — stops being representable, and `qualify_keys` drops its `Result` and becomes total.

### 2. Symbol keys under `#::` / `#::alias`

Supported, via a new `FormKind::AutoSymbol` — the symbol counterpart of `AutoKeyword`, with no surface syntax of its own. The reader still hands namespace resolution to whoever holds an `Env` rather than learning about `*ns*`.

```clojure
'#::{a 1}       ;; => {user/a 1}
'#::alias{a 1}  ;; => {resolved.alias/a 1}
#::{a 1}        ;; => Unable to resolve symbol: user/a   (as on the JVM)
```

### 3. Quote / IR divergence — and the duplication under it

Fixed as you described: `resolve_auto_forms` (in cljrs-builtins) qualifies every auto identifier in a form tree, and `quote`, the `'` sugar and `lower_arity` all call it, so no tier below sees an unresolved one. `'#::{:a 1}`, `#::{:a 1}` and `#::alias{:a 1}` now agree, verified on the tree-walker, the IR interpreter (`CLJRS_IR_THRESHOLD=1`) and the JIT.

The bigger thing: it turned out `macros::resolve_auto_kws` was already doing this job for macro arguments, over 8 of the `FormKind` variants, with `_ => clone()` for the rest. Variants added after it was written were silently skipped — so on current main, before this PR:

```clojure
(defmacro q [f] (list 'quote f))
(q #(vector ::k))   ;; cljrs: (fn* [] (vector :k))   JVM: (fn* [] (vector :my.app/k))
(q @(atom ::k))     ;; cljrs: (deref (atom :k))      JVM: (deref (atom :my.app/k))
(q [#?(:rust ::k)]) ;; cljrs: [:k]
```

No error, just the wrong keyword. `resolve_auto_forms` replaces it, so there is one walker instead of two and the extra positions come for free. Happy to split this into its own PR if you would rather keep #298 to the reader.

### 4. Whitespace before `{`

Skipped — spaces, tabs, newlines, commas. Comments are not, so `#:foo ;; c⏎{:a 1}` stays a read error, matching the JVM's "Namespaced map must specify a map".

### 5. `/` keys

`#:foo{/ 1}` has the symbol key `foo//` and `#:foo{:/ 1}` the keyword key `:foo//`. The old test pinned the opposite; it now pins the JVM behaviour.

### Cargo.lock

Updated.

### One more thing, not in this PR

`is_symbol_start` lived in `lexer.rs`, and `namespaced_map.rs` — a pure rewrite over already-parsed forms — needed it, which made the two modules mutually dependent. The predicates are total functions of one `char`, so they moved to a new `chars.rs` under both. No behaviour change; it is in this PR only because the prefix validation is what needed it.

### Testing

- The existing property file gains the prefix laws; a new lexer property file states that separators before `{` cannot change which prefix was read, and that a qualified or numeric prefix is always refused.
- A new interp property file states that no auto identifier survives resolution, at any nesting under any wrapper. Its counter matches `FormKind` exhaustively with no wildcard arm, so a variant added later fails to compile rather than going unwalked — which is the failure mode that produced finding 3.
- `cargo mutants` on `namespaced_map.rs` + `chars.rs`: 22 mutants, 16 caught, 4 unviable, 0 missed. Two timeouts, both mutating `is_symbol_char` into something that makes the lexer never terminate.
- Full workspace suite green.

Separately, `(meta (q ^{:x ::k} [1]))` is `nil` on cljrs and `{:x :my.app/k}` on the JVM — `form_to_value`'s `Meta` arm drops the metadata map. `resolve_auto_forms` does walk into it, so the `::k` inside is resolved correctly and then discarded. Out of scope here; I can open a separate issue or PR if you want it.

PR #314, opened 2026-08-11
feat(aot): audit what the wasm module leaves out, under the same policy

BuddhiLW:
Stacked on #293, which it does not change. This is the other half of your review: "would be good to enforce the flag under those combinations".

## The two backends fall short differently

`compile_file_to_wasm` embeds no source, so `SourceAudit` finds nothing there and #293 refuses the flag as unsupported. But the wasm path does not fully compile the program either. It drops what it cannot lower: a required namespace that fails lowering or codegen is skipped, and an entry form that `needs_interpreter` is filtered out of `__cljrs_main`. Both are left for an IR-interpreter tier that is not wired up yet.

So the module is opaque AND quietly missing behaviour, which is arguably worse than a leaky native binary. Today the only signal is one `eprintln!` per skipped namespace, and none at all for a dropped entry form.

## What this adds

`BundleAudit`, the wasm counterpart of `SourceAudit`, collecting `Omission`s during lowering. `compile_file_to_wasm` takes the SAME `OpacityPolicy` and applies the SAME `OpacityVerdict`, so `--require-fully-compiled` means one thing across backends: the artifact fully represents the program. Each backend reports its own way of falling short.

```
$ cljrs compile omit.cljrs -o omit.wasm --target wasm --require-fully-compiled
Error:   × --require-fully-compiled: 2 unit(s) of the program would be missing
  │ from the wasm module:
  │   • entry form (user): (defmulti route)
  │   • entry form (user): (defmethod route)
  │ The wasm backend drops what it cannot lower, leaving it for an IR-
  │ interpreter tier that is not wired up yet. Move the forms above out of the
  │ compiled unit, or drop --require-fully-compiled.
```

Nothing is written on rejection. `Report` stays the default and adds one summary line, so existing wasm builds are unchanged.

`--test` is untouched and still refused. `compile_test_harness` pushes every test namespace to `interpreted_bundled` unconditionally, so a strict policy there could never be satisfied; that rejection is the terminal answer rather than a placeholder. With wasm now audited, `resolve_opacity_policy` no longer takes a target, since the target stopped changing the decision.

## API

`compile_file_to_wasm` gains an `opacity` parameter. That is the only breaking signature change; the two existing `wasm_compile.rs` call sites pass `OpacityPolicy::Report` and behave as before.

## Tests

4 properties over `BundleAudit` mirroring the `SourceAudit` ones: completeness is exactly "recorded nothing", only the strict policy can reject, the report names every omission (it is the only thing telling a caller what to move), and monotonicity. 17 properties green in `source_leak_audit.rs`.

2 e2e wasm compiles in `aot_e2e.rs`: a `defmulti`/`defmethod` program is rejected and writes nothing, a plain `defn` program is accepted. CLI combination tests updated.

`cargo fmt` and `cargo clippy --tests` clean. Both crate READMEs updated in the same commit.

### DIFF
--- a/crates/cljrs-compiler/README.md
+++ b/crates/cljrs-compiler/README.md
@@ -338,6 +338,16 @@ impl SourceAudit {
 }
 
 pub enum OpacityPolicy { Report /* default */, RequireFullyCompiled }
+
+pub struct BundleAudit;                 // what the wasm module would omit
+impl BundleAudit {
+    pub fn of(omissions: impl IntoIterator<Item = Omission>) -> Self;
+    pub fn omissions(&self) -> &[Omission];
+    pub fn is_complete(&self) -> bool;
+    pub fn count(&self) -> usize;
+    pub fn verdict(&self, policy: OpacityPolicy) -> OpacityVerdict;
+}
+pub enum OmissionKind { Namespace, EntryForm }
 pub enum OpacityVerdict { Clean, Tolerated, Rejected }
 ```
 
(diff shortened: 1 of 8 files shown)


BuddhiLW:
Rebased onto main; the conflict is gone.

It was semantic rather than textual. #313 and `202f7d37` landed while this sat: `--target` became a closed `CompileTarget` enum, and the enum is now threaded through `resolve_opacity_policy` so it can refuse `--require-fully-compiled` for wasm — on the grounds that the wasm backend audits nothing.

This PR is what makes that no longer true, so the rebase keeps the enum everywhere and drops only the refusal. That leaves `target` unused, so the parameter comes out and `resolve_opacity_policy(require_fully_compiled, test)` is the signature again. Calling that out because it removes a parameter you threaded through one commit earlier — it is the consequence of the audit landing, not a revert of the refactor.

The unit test asserting the wasm refusal goes with it, and `test_combination_is_reported_before_wasm` no longer has a wasm arm to be reported before. What replaces both is `require_fully_compiled_is_accepted_on_the_wasm_path` in `tests/compile_flag_combinations.rs`, which drives the real CLI rather than the resolver in isolation.

`cargo fmt --check`, `clippy` on cljrs + cljrs-compiler, and the `main.rs` unit tests are clean locally; leaving the full suite to CI.

PR #317, opened 2026-08-12
ci: retry crates.io publish when rate-limited

csm:
The workspace has 33 publishable crates, which exceeds crates.io's
publish burst allowance (30 new versions of an existing crate, then 1
per minute; 5 for brand-new crates, then 1 per 10 minutes). A single
`cargo publish --workspace` therefore ends in a 429 almost every run,
with the tail of the workspace unpublished and a manual re-run needed.

Move the publish step into .github/scripts/publish-crates.sh, which
loops up to PUBLISH_MAX_ATTEMPTS (default 5) times:

- Before each attempt it asks crates.io which crates are already live at
  the target version and passes --exclude for them, so a partial
  success always makes forward progress and nothing is published twice.
- On a 429 it waits until the retry-after timestamp in the crates.io
  error message, clamped to [60s, PUBLISH_MAX_SLEEP]; when no timestamp
  can be parsed it falls back to attempt-scaled backoff
  (PUBLISH_BASE_SLEEP * attempt, default 300s, 600s, ...).
- On a non-rate-limit failure it waits only 60s, so a genuine build or
  auth error still fails reasonably quickly.
- The registry, not cargo's exit status, has the final word: if every
  crate is live the step succeeds even when cargo exited non-zero on the
  last one. Otherwise it fails listing exactly what is still missing.

Packages with `publish = false` (examples/rust-interop) are now filtered
out via cargo metadata's `publish` field instead of being probed against
crates.io, so they never count as outstanding work.

The publish job gains timeout-minutes: 120 — comfortably above the
worst-case ~50 minutes of accumulated backoff, well below the 6 h
default.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_013YHqTtQdDnj2gYASJy3Kya

### DIFF
--- a/.github/scripts/publish-crates.sh
+++ b/.github/scripts/publish-crates.sh
@@ -1,0 +1,173 @@
+#!/usr/bin/env bash
+#
+# Publish every publishable workspace crate to crates.io, retrying when
+# crates.io rate-limits us.
+#
+# Why this exists: the workspace has ~33 publishable crates and crates.io
+# throttles publishes (as of writing: new versions of an existing crate are a
+# burst of 30 then 1 per minute; brand-new crates are a burst of 5 then 1 per
+# 10 minutes).  A full workspace publish therefore reliably ends in
+#
+#     the remote server responded with an error (status 429 Too Many Requests):
+#     You have published too many crates ... Please try again after <timestamp>
+#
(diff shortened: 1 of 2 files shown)


csm:
Less important now that the crate count has been reduced.

PR #331, opened 2026-08-17
Gate the runtime's git dependency behind a `deps` feature

csm:
`cljrs-runtime` depended on `cljrs-project` unconditionally, and `GlobalEnv`
held `cljrs_project::vcs::TrustedKeys` as a struct field. Because it was a
field rather than an interface, nothing that linked the interpreter could
avoid gitoxide, rPGP and `ssh-key` — nor, through gix's http transport,
reqwest/hyper/rustls/aws-lc-rs/tokio/icu. An editor plugin, a sandboxed
evaluator or `cljrs-tx` linked a TLS stack to run Clojure.

Introduce `env::vcs::VcsProvider`: the four git operations the interpreter
actually needs (find repo root, read a file at a commit, verify a commit
signature, install trusted signers), as a trait object on `GlobalEnv`.

- `cljrs-runtime`'s new default-on `deps` feature installs the
  `cljrs-project`-backed `ProjectVcs` in `GlobalEnv::new`, so behaviour and
  every downstream crate are unchanged.
- With `--no-default-features` no provider is installed: source files are
  treated as living outside a repository, versioned resolution falls back to
  embedded (AOT) sources, and the gix/pgp/ssh-key tree is not compiled.
  303 -> 87 transitive crates for a runtime-only build.
- The provider field exists in both configurations, so `GlobalEnv`'s layout
  does not vary with the feature and feature unification cannot surprise an
  embedder. `set_vcs_provider` lets a host supply its own implementation, or
  pass `None` to strip git access from a sandbox.
- `check_commit_signature` now *fails* when verification is demanded but no
  provider exists, rather than silently accepting. Previously the wasm path
  passed unconditionally.

Split `cljrs-project`'s git support into two feature tiers so the weight
matches the use:

- `vcs` — local repository reads and signature verification. No network stack.
- `vcs-net` — `fetch_remote`, which is what selects gix's blocking http
  transport and with it reqwest/rustls/aws-lc-rs. Gitoxide routes every clone
  through that layer, so this gates local-path and `file://` fetches too.
  Only the `cljrs` binary takes it; the runtime never fetches, so even with
  `deps` on it links no HTTP/TLS stack (previously 0 of these crates were
  avoidable).

The http transport was selected by the workspace `gix` entry rather than by
gix's defaults, so `default-features = false` was already in place; moving
the transport feature to `vcs-net` is what actually drops it.

New `tests/vcs_provider.rs` covers the seam and passes both with and without
`deps`.

### DIFF
--- a/crates/cljrs-runtime/src/env/env.rs
+++ b/crates/cljrs-runtime/src/env/env.rs
@@ -684,8 +684,27 @@ impl GlobalEnv {
     /// `deps` feature supply its own git implementation, or a sandboxed host
     /// remove the default one (`None`) so no versioned resolution can reach
     /// the filesystem's git history.
+    ///
+    /// Drops every cached signature verdict: those were reached by the
+    /// outgoing provider, against its trust set and its view of the
+    /// repository, and say nothing about what the incoming one would decide
+    /// for the same `(repo, commit)`.  Keeping them would let a permissive
+    /// provider launder an approval for source a later provider serves.
     pub fn set_vcs_provider(&self, provider: Option<Arc<dyn crate::env::vcs::VcsProvider>>) {
+        // Neither this nor `check_commit_signature` ever holds the `vcs` and
+        // `sig_verify_cache` locks at the same time, and the two reach for
+        // them in opposite orders — keep it that way, or the pair becomes a
+        // lock-order inversion.
         *self.vcs.write().unwrap() = provider;
+        self.invalidate_signature_cache();
+    }
+
+    /// Forget every cached signature verdict, so the next
+    /// [`check_commit_signature`](Self::check_commit_signature) re-asks the
+    /// current provider.  Called whenever the thing that produced those
+    /// verdicts changes: the provider itself, or its trusted-key set.
+    pub fn invalidate_signature_cache(&self) {
+        self.sig_verify_cache.lock().unwrap().clear();
     }
 
     /// If `:verify-commit-signatures` is enabled, verify that `commit` inside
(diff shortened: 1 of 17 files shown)


crates/cljrs-runtime/src/env/env.rs
```diff
@@ -674,76 +673,74 @@ impl GlobalEnv {
...
+    pub fn vcs(&self) -> Option<Arc<dyn crate::env::vcs::VcsProvider>> {
+        self.vcs.read().unwrap().clone()
+    }
+
+    /// Replace the VCS backend.  Lets an embedder that built without the
+    /// `deps` feature supply its own git implementation, or a sandboxed host
+    /// remove the default one (`None`) so no versioned resolution can reach
+    /// the filesystem's git history.
+    pub fn set_vcs_provider(&self, provider: Option<Arc<dyn crate::env::vcs::VcsProvider>>) {
+        *self.vcs.write().unwrap() = provider;
```
csm, line 698: **[P2] Invalidate the signature cache when replacing the VCS provider**

`set_vcs_provider` replaces the provider but retains `sig_verify_cache`. After provider A approves a `(repo, commit)` pair, provider B can be installed and supply different source through `file_at_commit`; `check_commit_signature` returns the cached success without consulting B. Clear the cache when replacing the provider, or key it by a provider/trust generation and ensure verification and source reads use the same provider.

csm (commented):
One actionable issue found.

crates/cljrs-runtime/src/env/env.rs
csm: Fixed in 62eb30b — you're right, and the test I wrote asserted the stale-cache behaviour as if it were intended, which is how I missed it.

`set_vcs_provider` now clears `sig_verify_cache`, via a new public `invalidate_signature_cache`. I also call it from `load_trusted_signers`: replacing the key set invalidates verdicts reached under the old one for the same reason replacing the provider does. In the normal flow that runs at session start before anything is verified, so it costs nothing there.

I went with clearing rather than a generation counter — the cache is session-scoped and provider swaps happen at startup or sandbox teardown, so there's nothing worth preserving across one.

On lock ordering: `set_vcs_provider` reaches for `vcs` then `sig_verify_cache`, `check_commit_signature` the other way round. Neither holds both at once, so there's no inversion today; there's a comment on `set_vcs_provider` saying to keep it that way.

Test coverage replaced with a call-counting provider: `provider_swap_invalidates_cached_verdicts` (a pass is cached within one provider — one call for two checks — then a rejecting provider is swapped in and the check fails) and `reloading_trusted_signers_invalidates_cached_verdicts`. Both pass with and without `deps`; full suite 1183 passed, 0 failed.

---

PR #343, opened 2026-08-20
fix(aot): establish the entry namespace structurally, not as embedded source

BuddhiLW:
## What

The AOT entry namespace was established by embedding a source string into the
generated harness and evaluating it at startup. This replaces that with a
structural declaration: the `ns` form is recognised as data (name + `:require`
clauses), and the harness establishes the namespace directly.

## Why

Embedding source into a compiled artifact is a source leak — the point of AOT is
that the shipped binary does not carry, or need, the text it was built from. It
also means the entry namespace is established by a code path (the reader plus
the evaluator) that the rest of the AOT pipeline has already been careful to
avoid.

## Evidence

`crates/cljrs-compiler/tests/source_leak_audit.rs` and the `aot_e2e` cases
(`require_fully_compiled_rejects_embedded_source`,
`require_fully_compiled_accepts_plain_defn`).

---

Applies cleanly to current `main`.

### DIFF
--- a/crates/cljrs-compiler/README.md
+++ b/crates/cljrs-compiler/README.md
@@ -638,6 +638,34 @@ Pinned *versioned* sources (`mylib@<sha>`) are the exception — they still embe
 as interpreted builtin source, since they resolve through the separate
 versioned loader rather than the plain `require` path.
 
+**Structural entry namespace:** the *entry* file's own `ns` form is normally an
+interpreted form too, so it ships as readable text and `--require-fully-compiled`
+rejects the program. When that `ns` is the **only** interpreted form and carries
+no clause richer than `:require` — no `:import`, `:gen-class`, docstring or
+`:refer-clojure` — `structural_ns_form` recognises it and the harness establishes
+the namespace directly instead: `get_or_create_ns`, `refer_core`, `sync_star_ns`
+(so `(resolve '-main)` looks in the entry namespace rather than `user`), then one
+`load_ns` call per require. No source text is emitted, so a single-file program
+can pass the opacity gate.
+
+Require specs are parsed by `cljrs_runtime::interp::special::parse_require_spec_form`
+— the same function `eval_ns` uses — and emitted whole, alias and `:refer` and
+`@version` included. This path must not re-derive spec parsing: aliases are
+resolved at compile time by `qualify_aliases`, but `:refer` is not. A bare `f`
+lowers to `LoadGlobal(<entry-ns>, "f")` and is resolved through the namespace's
+refer table at *runtime*, and `rt_call` on a nil callee returns nil rather than
+erroring — so a dropped refer produces a program that runs and prints wrong
+answers. A dropped `@version` likewise routes a pinned require to the plain
+loader instead of the versioned one.
+
+The structural path is taken only when the `ns` is the *first* interpreted form;
+otherwise it is restored at the head of the preamble, which would reorder the
+program if anything interpreted had preceded it. Any other interpreted form
+present at all forces the textual preamble, since that form may depend on the
+aliases and refers the `ns` installs. Required namespaces still ship their own
+`ns`/`require` preambles as text, so multi-file programs remain rejected by
+`--require-fully-compiled` through the `NamespacePreamble` channel.
+
 ---
 
 ## Features
--- a/crates/cljrs-compiler/src/aot.rs
+++ b/crates/cljrs-compiler/src/aot.rs
@@ -816,17 +816,38 @@ pub fn compile_file(src_path: &Path, out_path: &Path, session: &CompileSession) 
     // The rest is AOT-compiled.
     let mut interpreted_source = String::new();
     let mut compilable = Vec::new();
+    // The entry `ns` form, held back: if it is the ONLY interpreted form, the
+    // harness can establish the namespace structurally and ship no text at all.
+    // Held back rather than dropped, because any OTHER interpreted form may
+    // depend on the aliases and refers this form installs.
+    let mut structural_ns: Option<(String, Vec<String>)> = None;
+    let mut ns_form_text = String::new();
     for (i, form) in expanded.iter().enumerate() {
         if needs_interpreter(&forms[i]) || expanded_needs_interpreter(form) {
             // Extract the original source text using span byte offsets.
             let span = &forms[i].span;
             let src_text = &source[span.start..span.end];
-            interpreted_source.push_str(src_text);
-            interpreted_source.push('\n');
+            match structural_ns_form(&forms[i]) {
+                Some(decl) if structural_ns.is_none() && ns_form_text.is_empty() => {
+                    structural_ns = Some(decl);
+                    ns_form_text.push_str(src_text);
+                    ns_form_text.push('\n');
+                }
+                _ => {
+                    interpreted_source.push_str(src_text);
+                    interpreted_source.push('\n');
+                }
+            }
         } else {
             let form = expand_anon_fns(form);
             compilable.push(qualify_aliases(&form, &env.current_ns, &env.globals));
         }
+    }
+    // Another form needs the interpreter, so the ns form must be evaluated the
+    // old way — its aliases have to exist before that form runs.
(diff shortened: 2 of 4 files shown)


csm (changes requested):
The direction is right, and the conservative classification (`:import`, `:gen-class`, docstring, `:refer-clojure` → keep the textual preamble) is the correct instinct. But the structural path drops require-spec information that the compiled body still depends on at runtime, and that is a confirmed, silent regression.

## Blocking

### 1. `:refer` is silently dropped — compiled code resolves referred symbols to `nil`

`structural_ns_form` inspects only the head of a require-spec vector (`crates/cljrs-compiler/src/aot.rs:1277`) and discards the rest, and the emitted spec hardcodes `alias: None, refer: RequireRefer::None` (`crates/cljrs-compiler/src/aot.rs:2144`).

Aliases are genuinely safe — `qualify_aliases` rewrote `b/foo` → `a.b/foo` at compile time. Refers are not. `lower_symbol` (`crates/cljrs-ir/src/lower/anf.rs:2817`) lowers a bare `f` to `LoadGlobal(<entry-ns>, "f")`, which `rt_lookup_global` resolves at runtime via `GlobalEnv::lookup_in_ns` — interns, *then refers* (`crates/cljrs-runtime/src/env/env.rs:368`). The structural path installs only the `clojure.core` refers, so the name is never found. `rt_call` on a nil callee returns nil rather than erroring, so the program prints wrong answers instead of failing.

Verified end to end with the existing `crates/cljrs-compiler/tests/aot_e2e.rs::test_multi_file_refer`:

```
--- multi_refer output mismatch ---
expected:
25
14
actual:
nil
nil
```

That test passes on `main` and fails on `da3d399`. It didn't show up in CI because it is behind the `aot_full_test` feature, which the workflow does not enable.

### 2. Versioned pins lose their commit

`mylib@abc1234` lexes as a *single* `Symbol` (`crates/cljrs-reader/src/lexer.rs:550`), so it lands in `requires` verbatim and is emitted as `RequireSpec { ns: "mylib@abc1234", version: None }`. `load_ns` routes to `load_versioned_ns` only when `spec.version.is_some()` (`crates/cljrs-runtime/src/env/loader.rs:24`), so a pinned require in the entry `ns` now takes the plain loader path instead. `parse_require_spec_form` gets this right via `Symbol::parse`; the new path re-derives spec parsing and diverges from it.

**Suggested fix for both:** don't re-derive spec parsing. Make `cljrs_runtime::interp::special::parse_require_spec_form` public, have `structural_ns_form` return `(String, Vec<RequireSpec>)`, and emit the full spec — ns, version, alias, refer — into the harness. That keeps this path in sync with `eval_ns` by construction rather than by re-derivation. Installing aliases costs nothing and closes the remaining hole for anything that resolves dynamically at runtime.

### 3. CI is red

`cargo fmt --check` fails on two hunks (`aot.rs:2147` and `aot.rs:3152`).

## Worth addressing

**4.** `.expect("require failed at startup")` (`aot.rs:2145`) — every other startup failure in the harness prints to stderr and `exit(1)`. A missing dependency here panics and unwinds through `resume_unwind` in `main`.

**5.** Reordering when the `ns` is not the first interpreted form: `interpreted_source.insert_str(0, &ns_form_text)` (`aot.rs:862`) puts the `ns` ahead of any interpreted form that preceded it in the file. Rare, but silent. Recording the ns form's index and taking the structural path only when it is the first interpreted form removes the case entirely — and would also let `ns_form_text` become an `Option`, making the redundant `ns_form_text.is_empty()` guard at `aot.rs:843` unnecessary.

**6.** `refer_all(ns, "clojure.core")` (`aot.rs:2133`) vs `eval_ns`'s `refer_core(ns)`. Equivalent today, since the structural path rejects `:refer-clojure` so no filter can exist. But `refer_core` is the API that *means* "the automatic core refer", it stays correct if that interaction changes, and `eval_ns` also skips it when the namespace is `clojure.core` itself.

## Nits

**7.** No e2e coverage of the new harness code. The four unit tests cover `structural_ns_form`'s classification, but nothing compiles and runs a binary through the structural path — and the generated Rust is only type-checked when someone actually builds a program. The two `aot_e2e` cases named in the description (`require_fully_compiled_rejects_embedded_source`, `require_fully_compiled_accepts_plain_defn`) predate this PR and have no `ns` form, so neither exercises it. A case that compiles `(ns m) (defn -main [] ...)` under `RequireFullyCompiled` and asserts the binary's output would have caught #1's sibling cases too.

**8.** `crates/cljrs-compiler/README.md` documents the entry preamble partition but not the structural entry-ns path; per `CLAUDE.md` that belongs in the same commit.

One scope note, not a defect: this only closes the source leak for single-file entry programs — required namespaces still ship their own `ns`/`require` preambles as text, so multi-file programs remain rejected by `--require-fully-compiled` through the `NamespacePreamble` channel.

---

csm (commented):
Re-reviewed at `a7fc0bb`. Every blocking item is fixed, and fixed the right way — routing through `parse_require_spec_form` rather than patching the two dropped fields means this path can't drift from `eval_ns` again. Only nits remain.

## Verified

- **`:refer`** — `crates/cljrs-compiler/tests/aot_e2e.rs::test_multi_file_refer` passes again (it printed `nil / nil` at `da3d399`). The new `structural_entry_ns_keeps_referred_names_resolvable` covers the same ground under the gate, and both new e2e tests run the binary rather than only compiling it, which is the property that was missing.
- **`@version`** — carried through `render_require_spec`, with `a_require_spec_keeps_its_refer_and_version` and `a_rendered_spec_carries_every_field` pinning it.
- **`cargo fmt --check`** — clean; CI is green on both jobs.
- **Reordering** — the `interpreted_source.is_empty()` guard is the right shape: the fallback's `insert_str(0, ..)` is now correct by construction rather than by convention.
- **Startup require failure**, **`refer_core`** (including the `clojure.core` self-skip, decided at compile time since the namespace is known — nice), **`Option<String>`**, **compiler README** — all addressed.

`HarnessInputs` wasn't asked for and is a good call; `build_harness` was past the point where positional arguments carry meaning.

## Nits

**1. `render_require_spec` doesn't actually get the exhaustiveness it claims.** The commit message says "the generated struct literal is exhaustive, so a field added later becomes a compile error rather than another silent None", and the doc comment says "every field is carried". That holds for the `RequireRefer` variants — `match &spec.refer` will break on a new variant — but not for `RequireSpec`'s fields: the function reads them individually into a `format!`, so a fifth field added tomorrow compiles fine here and silently never reaches the harness. Destructuring makes the claim true:

```rust
let RequireSpec { ns, version, alias, refer } = spec;
```

**2. A reader-conditional require spec makes the whole `ns` non-structural.** `eval_ns` runs `expand_reader_conds(&items[1..])` before parsing (`crates/cljrs-runtime/src/interp/special.rs:1650`); `structural_ns_form` passes each spec form straight to `parse_require_spec_form`, which rejects a bare `ReaderCond` with "require spec must be a symbol or vector", so `.ok()?` declines the whole form. It fails safe — textual preamble, gate refuses, no wrong answers — but it means `(ns m (:require #?(:rust [a] :clj [b])))` can never pass `--require-fully-compiled`, while `(ns m (:require [#?(:rust a :clj b) :as x]))` can, since `spec_element` handles conditionals *inside* a vector spec. Given `.cljc` and `#?(...)` are core to the project, the same one-line `expand_reader_conds` call would close the gap and remove the inconsistency.

**3. `crates/cljrs-runtime/README.md` wasn't updated.** `parse_require_spec_form` is newly public API of *that* crate, and `CLAUDE.md` asks for the README in the same commit. The "`special.rs` notes" section already describes `spec_element` and require-spec conditional handling, so it's a one-sentence addition in an existing paragraph.

Separately, and not this PR's problem: the `aot_e2e` suite is behind the `aot_full_test` feature, which `cargo test --workspace` in CI doesn't enable — so the regression this round fixed would not have been caught by CI either before or after. Worth a follow-up issue to run that suite somewhere (nightly, if it's too slow for per-PR).

---

csm (approved):

csm:
Note that the `aot_full_test` is not enabled in CI because it takes far too long to run for GitHub actions to handle.

PR #344, opened 2026-08-20
fix(reader): parse special forms through metadata

BuddhiLW:
## What

`^m form` now has the structural shape of `form`. `Form` gains `unmeta` /
`peel_meta` plus `as_symbol` / `as_keyword` / `as_string` / `as_list` /
`as_vector` / `as_map` views that report the shape underneath any annotation,
and the interpreter's special forms and the IR lowerer ask those instead of
matching `FormKind` directly.

## Why

An annotation is metadata, not a change of shape — but every site that matched
`FormKind::Vector(..)` directly disagreed. `(defrecord R [^String x])` and
`(defn f ^String [s] s)` failed to parse, both of which are ordinary JVM
Clojure.

## Evidence

`crates/cljrs-runtime/tests/meta_transparency.rs`: each special form asserted
equal with and without an annotation at every structural site.

---

**Ported, not cherry-picked.** The original work predates the four-crate
consolidation into `cljrs-runtime`, and `main` has since added
reader-conditional/pair expansion (`expand_pairs`, `expand_reader_conds_cow`)
inside the very matches this changes. Each of those keeps `main`'s expansion and
only swaps the match for the structural view. One hunk of the original is
dropped entirely: `main`'s own fix for stacked metadata at a `def` name
supersedes it, so `extract_def_name` is left as `main` has it.

Base of a five-PR stack: #344 → #345 → #346 → #347 → #348. Each is cut against `main`, so until this one merges the later PRs' diffs include the commits below them. Merging in that order keeps every diff minimal.

### DIFF
--- a/crates/cljrs-ir/src/lower/anf.rs
+++ b/crates/cljrs-ir/src/lower/anf.rs
@@ -817,7 +817,7 @@ fn lower_let(ctx: &mut LowerCtx, args: &[Form]) -> R {
             "let requires a binding vector".into(),
         ));
     }
-    let FormKind::Vector(bindings) = &args[0].kind else {
+    let Some(bindings) = args[0].as_vector() else {
         return Err(LowerError::MalformedSpecialForm(
             "let bindings must be a vector".into(),
         ));
@@ -855,7 +855,7 @@ fn lower_loop(ctx: &mut LowerCtx, args: &[Form]) -> R {
             "loop requires a binding vector".into(),
         ));
     }
-    let FormKind::Vector(bindings) = &args[0].kind else {
+    let Some(bindings) = args[0].as_vector() else {
         return Err(LowerError::MalformedSpecialForm(
             "loop bindings must be a vector".into(),
         ));
@@ -963,21 +963,10 @@ fn lower_def(ctx: &mut LowerCtx, args: &[Form]) -> R {
         ));
     }
     // Name may carry metadata: `(def ^:private x 1)`.
-    let name_str: Arc<str> = match &args[0].kind {
-        FormKind::Symbol(s) => Arc::from(s.as_str()),
-        FormKind::Meta(_, inner) => {
-            let FormKind::Symbol(s) = &inner.kind else {
-                return Err(LowerError::MalformedSpecialForm(
-                    "def name must be a symbol".into(),
-                ));
-            };
-            Arc::from(s.as_str())
-        }
-        _ => {
-            return Err(LowerError::MalformedSpecialForm(
-                "def name must be a symbol".into(),
-            ));
-        }
+    let Some(name_str) = args[0].as_symbol().map(Arc::<str>::from) else {
+        return Err(LowerError::MalformedSpecialForm(
+            "def name must be a symbol".into(),
+        ));
     };
     let ns = ctx.ns().clone();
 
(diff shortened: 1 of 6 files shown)


csm (commented):
## Resolving the conflict with `main`

I reproduced the merge locally. It's a single hunk, in `crates/cljrs-ir/src/lower/anf.rs` inside `lower_def`:

```rust
<<<<<<< HEAD
    // Name may carry metadata, possibly stacked: `(def ^:private ^:dynamic x 1)`.
    let FormKind::Symbol(s) = &peel_meta(&args[0]).kind else {
=======
    // Name may carry metadata: `(def ^:private x 1)`.
    let Some(name_str) = args[0].as_symbol().map(Arc::<str>::from) else {
>>>>>>> (this PR)
```

### The two sides are behaviorally identical

`main` and this PR independently fixed the same stacked-metadata-on-a-`def`-name bug, with two different helpers. Compare the loops:

```rust
// main: anf.rs (free fn)                              // this PR: cljrs-reader/src/form.rs (method)
fn peel_meta(form: &Form) -> &Form {                   pub fn unmeta(&self) -> &Form {
    let mut inner = form;                                  let mut form = self;
    while let FormKind::Meta(_, next) = &inner.kind {      while let FormKind::Meta(_, inner) = &form.kind {
        inner = next;                                          form = inner;
    }                                                      }
    inner                                                  form
}                                                      }
```

Both strip *every* `Meta` layer; `as_symbol()` is `unmeta()` plus a `Symbol` match. Same accepted inputs, same rejections, same error, same `Arc<str>`. So the choice is a maintenance question, not a behavioral one — nothing downstream can observe it.

### Two things that aren't visible in the conflict markers

**1. `--theirs` doesn't compile.** Two lines below the conflict, in shared context, is `main`'s:

```rust
let name_str: Arc<str> = Arc::from(s.as_str());
```

This PR deletes that line (it binds `name_str` in the let-else); `main` needs it. Take this PR's side wholesale and `s` is undefined. Either resolution requires a hand-edit *outside* the marked region.

**2. `peel_meta` ends up meaning two different things in one file.** After the merge, `anf.rs` holds `main`'s free `fn peel_meta(&Form) -> &Form` alongside calls to this PR's new method `Form::peel_meta() -> (Vec<&Form>, &Form)` (at `lower_defn` and `lower_fn_arity`). Different return types, same name, same file. It compiles — call syntax disambiguates — so nothing will flag it, but it's a genuine trip hazard. This is the only lasting consequence of the choice, and it points at keeping this PR's side.

### Suggested resolution

Take this PR's side, drop the orphaned line, and collapse the duplicate helper. `main`'s free `peel_meta` has exactly one other caller (`def_name`), so converting it retires the shadowing name entirely:

```rust
// lower_def — keep main's comment, it documents the stacked case better
    // Name may carry metadata, possibly stacked: `(def ^:private ^:dynamic x 1)`.
    let Some(name_str) = args[0].as_symbol().map(Arc::<str>::from) else {
        return Err(LowerError::MalformedSpecialForm(
            "def name must be a symbol".into(),
        ));
    };
    let ns = ctx.ns().clone();     // <- main's `let name_str: Arc<str> = ...` line deleted

// def_name — was `&peel_meta(form).kind`
    let FormKind::Symbol(s) = &form.unmeta().kind else {

// and delete the now-unused free `fn peel_meta`
```

Worth keeping `main`'s comment wording over this PR's — `(def ^:private x 1)` understates what the code handles.

### Verified

Applied exactly the above on a probe branch off `main`:

- `cargo check -p cljrs-ir` — clean
- `cargo clippy -p cljrs-ir` — no warnings
- `cargo test -p cljrs-ir` — 24/24, including `stacked_metadata_on_a_def_name_still_shadows` (`crates/cljrs-ir/tests/core_shadowing.rs`), `main`'s guard for its half of the fix
- `cargo test -p cljrs-runtime --test meta_transparency` — 21/21, this PR's guard for its half

Both sides' regression tests pass under the single resolution, which is the concrete confirmation that the choice is free.

### One note on the PR description

> One hunk of the original is dropped entirely: `main`'s own fix for stacked metadata at a `def` name supersedes it, so `extract_def_name` is left as `main` has it.

That's right about the `def_name` *helper*, which this PR correctly leaves alone — but `main` also changed `lower_def` itself, and that's where the conflict landed. The stated intent (defer to `main` on stacked-meta `def` names) holds either way, since the two are equivalent.

---

csm:
See #358 for fully merged version to main.

PR #345, opened 2026-08-20
fix(reader): keep reader metadata on the value

BuddhiLW:
## What

`form_to_value` dropped the `Meta` arm's annotation, so `(meta '^{:x 1} [1])`
was nil where the JVM answers `{:x 1}`, and any macro inspecting `(meta arg)`
saw nothing. Both the quoted and the evaluated position now attach it.

Inside `quote` the annotation is data (`'^{:x (+ 1 2)} [1]` keeps the
unevaluated list); outside it the meta form is evaluated. Shorthands expand as
the reader does (`^:kw` → `{:kw true}`, `^Sym` → `{:tag Sym}`), stacked
annotations merge with the outer one winning, and values that cannot carry
metadata drop it rather than growing a wrapper.

## Three latent faults this surfaced

Making a previously-dropped value flow exposed code that relied on its absence:

- `->` threads with `(with-meta … (meta form))`, and `(meta form)` is nil for an
  unannotated form. A nil annotation must carry nothing, so `(with-meta x nil)`
  now clears metadata instead of storing a nil-meta wrapper — which was neither
  `identical?` to itself after a clone nor reported by `type`.
- `type_tag_of` did not look through a metadata wrapper, so `(type ^{:a 1} {})`
  answered Object.
- `^Sym` produced `{:tag "Sym"}`; the JVM produces the symbol.

## Evidence

`crates/cljrs-runtime/tests/form_metadata.rs`, plus the vendored suite —
`conj`, `empty`, `identical?`, `random-uuid` and `sort-by` all exercise the
three faults above and were the ones that caught them.

---

metadata-transparency change; please merge that one first.

Stacked on #344 — please merge that first.


csm (commented):
## Review — `fix(reader): keep reader metadata on the value`

Note on stacking: #344 landed as #358, so the "merge that first" line in the description is stale — this branch's merge base is already `445179a` (current `main`) and `git diff main...HEAD` is exactly the 8 files, nothing from `main` reverted.

Verified locally at `a1e945d`: `cargo build --workspace` clean, `cargo test -p cljrs-runtime` green across all 39 targets (including the `meta_transparency` oracle from #358). Everything below is behaviour the suite doesn't reach; every repro is `cljrs eval` against a debug build of this head.

The semantics you picked are the right ones — quoted-is-data / evaluated-is-evaluated, outer-wins merge, `IObj`-style drop. Two things block merge, though.

### Blocking

**1. Only the tree-walk tier attaches the annotation; the IR tier drops it (`crates/cljrs-ir/src/lower/anf.rs:453`).**

```
$ cljrs eval '(defn f [] (meta ^{:a 1} [1])) (println "cold:" (f)) (dotimes [_ 5000] (f)) (println "hot:" (f))'
cold: {:a 1}
hot: nil
```

`ExecutionMode::Tiered` is the default and what the CLI uses, so the same expression answers `{:a 1}` until the fn is promoted and `nil` after — a result that depends on how hot the code got. `lower_form`'s `FormKind::Meta` arm only sniffs for `^:async (fn …)` and otherwise recurses into `inner`, discarding the annotation, and AOT lowers through the same path, so compiled binaries get the `nil` unconditionally. The quoted position is fine in every tier because `form_to_value` is shared; it's specifically the new `interp/eval.rs:145` arm that has no IR counterpart.

`tests/form_metadata.rs:15` pins `ExecutionMode::TreeWalk`, which is why this is invisible. A `Tiered` variant of `quote_keeps_the_annotation_as_data` plus one evaluated-position case would have caught it.

**2. Metadata on symbols becomes routine, and the symbol accessors aren't meta-transparent.**

```
$ cljrs eval "(name ^{:a 1} 'x)"
Error: wrong type: expected named, got symbol
```

That error is self-contradictory because `type_name` unwraps the wrapper and the matcher doesn't. `builtin_name` (`builtins.rs:6319`) and `builtin_namespace` (`:6331`) match `&args[0]` directly, and so do ~34 `*_q` predicates — `(symbol? ^{:a 1} 'x)`, `(seq? ^{:a 1} '(1 2))`, `(fn? ^{:a 1} (fn [] 1))` all answer `false`, while `vector?`/`map?` do call `unwrap_meta()` and answer `true`.

It bites the ordinary name-mangling macro idiom:

```
$ cljrs eval "(defmacro dm [n] (list 'def (symbol (str \"make-\" (name n))) 1)) (dm ^:private thing) make-thing"
Error: wrong type: expected named, got symbol
```

All of these worked before this change only because the annotation was thrown away. This is the same class as the three latent faults the description calls out — the symbol family is the one still open. `supports_meta` has to keep `Value::Symbol` (the quoted-path tests depend on it), so the fix belongs in the accessors and predicates.

### Should fix

**3. `type_tag_matches` wasn't updated alongside `type_tag_of`** (`env/apply.rs:24` vs `:63`). Its own doc comment and `crates/cljrs-runtime/README.md:531` both say it "must agree exactly with `type_tag_of`". It no longer does: for `WithMeta(Vector, …)` the miss path caches `type_tag_of` → `"Vector"`, then the fast path in `rt_call_ic` re-derives `"Object"`, so every metadata-carrying dispatch value is a permanent inline-cache miss that re-resolves and rewrites the entry under lock on each call. Same one-line `unwrap_meta()` in both the outer and inner match.

**4. `value_to_form` re-evaluates the annotation** (`interp/macros.rs:350`). Rebuilding meta as a `FormKind::Meta` means the map is analysed again at `interp/eval.rs:152`:

```
$ cljrs eval "(defmacro m [] (with-meta '(vector 1) {:tag 'String})) (m)"
Error: Unable to resolve symbol: String
$ cljrs eval "(defmacro m [] (with-meta '(vector 1) {:arglists '([x])})) (m)"
Error: Unable to resolve symbol: x
```

Both returned `[1]` before. Metadata that reached `value_to_form` is already data — quoting the meta child (or otherwise marking it as not-to-be-analysed) keeps the round trip lossless without giving up the macro case the new test covers.

**5. The two shorthand tables already disagree.** `meta_form_to_value` (`form.rs:254`) handles `Str`, and auto-keywords reach it pre-resolved via `resolve_auto_forms`; `compile_meta_form` (`special.rs:155`) has neither, so the evaluated position yields a non-map:

```
(meta '^"String" [1])  ;=> {:tag "String"}     (meta ^"String" [1])  ;=> "String"
(meta '^::foo [1])     ;=> {:user/foo true}    (meta ^::foo [1])     ;=> :user/foo
```

Worth folding into one expander parameterised on whether the general case evaluates, so they can't drift again. Related: any non-map annotation is now attached verbatim — `(meta ^42 [1])` → `42` — where the JVM reader rejects it outright.

### Nits

- `supports_meta` omits `Queue`, `Macro`, `BoundFn` and `NativeFunction`, all `IObj` on the JVM.
- `eval.rs:149` tests `supports_meta` before evaluating the annotation, so `^{:x (f)} 42` skips the side effect while `^{:x (f)} [1]` runs it.
- Per CLAUDE.md the crate README should list the newly public `supports_meta`, `merge_meta_values` and `compile_meta_form` with signatures; right now they're only named in the prose paragraph.
- Adjacent and pre-existing, not caused here but squarely in this area: `syntax_quote.rs` has no `FormKind::Meta` arm, so annotated forms fall to `_other => form_to_value(form)` (`:165`) and skip auto-gensym — ```(let [^long y# 1] y#)`` fails with "Unable to resolve symbol: y__0__auto__" on `main` too.

### What's right

- `^Sym` → `{:tag Sym}` as a symbol matches the JVM, and no in-repo consumer reads `:tag` off a metadata *value* (`special.rs:494` and `anf.rs:685` both read the `Form`), so the string → symbol switch is safe.
- `(with-meta x nil)` clearing rather than storing a nil-meta wrapper, `type_tag_of` unwrapping, outer-wins merge order, and the quoted-vs-evaluated split all look right, and the test file documents the reasoning well.

---

crates/cljrs-runtime/src/interp/eval.rs
```diff
@@ -142,9 +142,22 @@ pub fn eval(form: &Form, env: &mut Env) -> EvalResult {
...
+            }
+            let m = crate::interp::special::compile_meta_form(meta, env)?;
+            if matches!(m, Value::Nil) {
+                return Ok(value);
+            }
+            let merged = match value.get_meta() {
+                Some(existing) => crate::builtins::form::merge_meta_values(existing, &m),
+                None => m,
+            };
+            Ok(value.with_meta(merged))
```
csm: **Blocking — this arm has no IR counterpart, so the result depends on how hot the code got.**

`cljrs-ir/src/lower/anf.rs:453` handles `FormKind::Meta` by sniffing for `^:async (fn …)` and otherwise recursing into `inner`, discarding the annotation. `ExecutionMode::Tiered` is the default and what the CLI uses, so once a fn is promoted this arm stops running:

```
$ cljrs eval '(defn f [] (meta ^{:a 1} [1])) (println "cold:" (f)) (dotimes [_ 5000] (f)) (println "hot:" (f))'
cold: {:a 1}
hot: nil
```

AOT lowers through the same path, so compiled binaries get the `nil` unconditionally. The quoted position is consistent across tiers because `form_to_value` is shared — it's only the evaluated position that diverges.

Two smaller notes on this arm while it's being touched:

- Line 149 tests `supports_meta` *before* evaluating the annotation, so `^{:x (f)} 42` skips the side effect while `^{:x (f)} [1]` runs it.
- Line 152 is where the metadata a macro attached gets re-analysed — see the comment on `macros.rs:350`.

---

crates/cljrs-runtime/src/builtins/form.rs
```diff
@@ -222,6 +222,55 @@ pub fn resolve_auto_forms(form: &Form, env: &crate::env::env::Env) -> EvalResult
...
+            | Value::Vector(_)
+            | Value::Map(_)
+            | Value::Set(_)
+            | Value::Symbol(_)
+            | Value::Cons(_)
+            | Value::LazySeq(_)
+            | Value::Fn(_)
+            | Value::TypeInstance(_)
+    )
+}
```
csm, line 246: **Blocking — keeping `Value::Symbol` here (correctly) makes metadata-wrapped symbols routine, and the symbol accessors aren't meta-transparent.**

```
$ cljrs eval "(name ^{:a 1} 'x)"
Error: wrong type: expected named, got symbol
```

The error contradicts itself because `type_name` unwraps the wrapper and the matcher doesn't. `builtin_name` (`builtins.rs:6319`) and `builtin_namespace` (`:6331`) match `&args[0]` directly, and so do ~34 `*_q` predicates — `(symbol? ^{:a 1} 'x)`, `(seq? ^{:a 1} '(1 2))` and `(fn? ^{:a 1} (fn [] 1))` all answer `false`, while `vector?`/`map?` do call `unwrap_meta()` and answer `true`.

It bites the ordinary name-mangling macro idiom:

```
$ cljrs eval "(defmacro dm [n] (list 'def (symbol (str \"make-\" (name n))) 1)) (dm ^:private thing) make-thing"
Error: wrong type: expected named, got symbol
```

All of these worked before only because the annotation was dropped — same class as the three latent faults in the description, and the symbol family is the one still open. Dropping `Value::Symbol` from this list isn't the fix (the quoted-path tests depend on it); the accessors and predicates are.

Separately, this list omits `Queue`, `Macro`, `BoundFn` and `NativeFunction`, all `IObj` on the JVM.

---

crates/cljrs-runtime/src/env/apply.rs
```diff
@@ -21,7 +21,7 @@ fn check_arity(arity: &Arity, argc: usize, name: &str) -> EvalResult<()> {
 
 /// Return the canonical type tag for a value (used by protocol dispatch).
 pub fn type_tag_of(val: &Value) -> Arc<str> {
-    match val {
+    match val.unwrap_meta() {
```
csm, line 24: This unwrap is right, but its companion wasn't updated with it. `type_tag_matches` (`:63`) still matches `val` directly, and both its own doc comment and `crates/cljrs-runtime/README.md:531` state it "must agree exactly with `type_tag_of`".

It no longer does. For `WithMeta(Vector, …)` the miss path in `rt_call_ic` (`cljrs-compiler/src/rt_abi.rs:4841`) caches `type_tag_of` → `"Vector"`, then the fast path re-derives `"Object"` and rejects its own cached entry — so every metadata-carrying dispatch value is a permanent inline-cache miss that re-resolves and rewrites the entry under lock on each call. Dispatch stays correct; the cache just never hits.

Same one-line `unwrap_meta()` fixes it, in both the outer match and the inner one.

---

crates/cljrs-runtime/src/interp/macros.rs
```diff
@@ -340,9 +340,15 @@ pub fn value_to_form(val: &Value, span: Span) -> EvalResult<Form> {
...
-            return value_to_form(inner, span);
+        // WithMeta: keep the annotation as a `^meta` wrapper so metadata a
+        // macro received (or attached) survives back into the AST.
+        Value::WithMeta(inner, meta) => {
+            if matches!(**meta, Value::Nil) {
+                return value_to_form(inner, span);
+            }
+            let inner_form = value_to_form(inner, span.clone())?;
+            let meta_form = value_to_form(meta, span.clone())?;
+            FormKind::Meta(Box::new(meta_form), Box::new(inner_form))
```
csm: Rebuilding the annotation as a `FormKind::Meta` means it gets *analysed again* at `eval.rs:152`, but metadata that reached `value_to_form` is already data:

```
$ cljrs eval "(defmacro m [] (with-meta '(vector 1) {:tag 'String})) (m)"
Error: Unable to resolve symbol: String
$ cljrs eval "(defmacro m [] (with-meta '(vector 1) {:arglists '([x])})) (m)"
Error: Unable to resolve symbol: x
```

Both returned `[1]` before this change, and `:tag`/`:arglists` are exactly the metadata macros carry. Quoting the meta child — or otherwise marking it as not-to-be-analysed — keeps the round trip lossless without giving up `metadata_survives_a_macro_round_trip`.

---

crates/cljrs-runtime/src/interp/special.rs
```diff
@@ -152,18 +152,18 @@ fn extract_def_name(form: &Form, env: &mut Env) -> EvalResult<(String, Option<Va
...
         FormKind::Symbol(s) => {
-            // ^TypeHint  →  {:tag "TypeHint"}
+            // ^TypeHint  →  {:tag TypeHint}
             let m = MapValue::empty().assoc(
                 Value::keyword(Keyword::parse("tag")),
-                Value::string(s.clone()),
+                Value::symbol(cljrs_value::Symbol::parse(s)),
             );
             Ok(Value::Map(m))
         }
```
csm: There are now two shorthand expanders, and they already disagree. `meta_form_to_value` (`form.rs:254`) has a `Str` arm and receives auto-keywords pre-resolved via `resolve_auto_forms`; this one has neither, so its `_ => eval(…)` fallback produces something that isn't a map:

```
(meta '^"String" [1])  ;=> {:tag "String"}     (meta ^"String" [1])  ;=> "String"
(meta '^::foo [1])     ;=> {:user/foo true}    (meta ^::foo [1])     ;=> :user/foo
```

The quoted column is right in both rows. Worth folding the two into one expander parameterised on whether the general case evaluates, so the shorthand table lives in one place and can't drift again.

Related, and true of both: a non-map annotation is now attached verbatim — `(meta ^42 [1])` → `42` — where the JVM reader rejects it outright.

The `{:tag "TypeHint"}` → `{:tag TypeHint}` change itself is correct and safe: `special.rs:494` and `anf.rs:685` both read `:tag` off the `Form`, and no in-repo consumer reads it off a metadata *value*.

---

crates/cljrs-runtime/tests/form_metadata.rs
```diff
@@ -0,0 +1,132 @@
...
+
+use std::sync::Arc;
+
+use cljrs_reader::Parser;
+use cljrs_runtime::env::env::{Env, GlobalEnv};
+use cljrs_value::Value;
+
+fn make_env() -> (Arc<GlobalEnv>, Env) {
+    let globals = cljrs_runtime::Runtime::builder()
+        .execution_mode(cljrs_runtime::ExecutionMode::TreeWalk)
```
csm, line 15: Pinning `TreeWalk` is what hides the tier divergence flagged on `eval.rs:145` — the whole file passes here and the evaluated-position assertions go red under the default `Tiered` mode once a fn is promoted.

A second `make_env` variant on `ExecutionMode::Tiered`, with the evaluated-position cases (`shorthand_annotations_expand`, `evaluated_annotations_see_the_enclosing_scope`, `type_sees_through_an_annotation`) run inside a fn called past the promotion threshold, would pin the behaviour in both tiers. The quoted cases already agree across tiers since `form_to_value` is shared.

---

crates/cljrs-runtime/README.md
```diff
@@ -646,6 +646,16 @@ Callers phrase `OddArity` in their own words (`map literal must have an even
...
+### Reader metadata on values
+
+`form_to_value` attaches a `^meta` annotation to the value it denotes, so
+`(meta '^{:x 1} [1])` answers `{:x 1}` as on the JVM. Shorthands expand as the
+reader does (`^:kw` → `{:kw true}`, `^Sym` → `{:tag Sym}`), stacked
+annotations merge with the outer one winning (`merge_meta_values`), and values
+that cannot carry metadata (`supports_meta` — the JVM's `IObj`) drop it rather
+than growing a wrapper. A nil annotation carries nothing, and
+`(with-meta x nil)` clears metadata instead of storing a nil-meta wrapper.
+
```
csm, line 699: The prose is accurate and reads well, but CLAUDE.md asks the crate README to carry the public API with signatures. `supports_meta` and `merge_meta_values` are new `pub fn`s and `compile_meta_form` was just widened from private to `pub` — right now they're only named mid-sentence. Adding them to the fenced signature block just above (the one ending in `form_to_value`) would keep this section complete.

Worth noting here too that `form_to_value` is the shared path, so this paragraph describes behaviour that holds in every tier — whereas the evaluated `^m expr` position currently only holds under `TreeWalk` (see the comment on `eval.rs:145`). If that divergence is fixed rather than documented, this section stays true as written.

---

csm (commented):
Line-anchored versions of the findings from my summary review above — same seven items, each on the code it applies to, with the repro next to the line that produces it.

Order of severity: `eval.rs:145` (tier divergence) and `form.rs:229` (symbol accessors) are the two blockers; `apply.rs:24`, `macros.rs:350` and `special.rs:155` are should-fixes; `tests/form_metadata.rs:15` and the README note are follow-ups.

Nothing new since the summary — no need to read both.

---

BuddhiLW:
Both blockers fixed, plus all three should-fixes and every nit. Two commits: `a6f14a1` (the rule + transparency), `d7c1eb5` (a `lower_quote` gap the first one made reachable).

## 1. Tier divergence

You're right that `anf.rs:453` has no IR counterpart. Before writing one I checked the JVM, and the tree-walk arm was over-attaching:

```
(meta ^{:a 1} (list 1))          ;=> nil
(meta ^{:a 1} x)                 ;=> nil
(meta ^{:a 1} (if true [1] [2])) ;=> nil
(meta ^{:a 1} [1])               ;=> {:a 1}
```

An evaluated-position annotation becomes runtime metadata only on a form that *constructs* an `IObj` (collection literal or `fn`). Everywhere else it's a hint and isn't evaluated at all. This branch answered `{:a 1}` for all four.

So the fix is one predicate, `Form::takes_runtime_meta`, in **cljrs-reader**, because `cljrs-ir` can't depend on `cljrs-runtime`. Both `interp::eval` and `lower::anf` ask it, so they can't disagree.

Narrowing makes the IR side cheap: lowering emits `meta` -> `merge` -> `with-meta` via the ordinary `LoadGlobal`+`Call` path, so no new `Inst`. `merge` rather than plain `with-meta` is what keeps both layers of `^:a ^:b [1]` alive. A type hint now costs nothing: `(f ^long n)` emits no attach.

Your repro:

```
$ cljrs eval '(defn f [] (meta ^{:a 1} [1])) (println "cold:" (f)) (dotimes [_ 5000] (f)) (println "hot:" (f))'
cold: {:a 1}
hot: {:a 1}
```

AOT binary, the tier with no cold path to mask it:

```
$ cljrs compile probe.cljrs -o probe && ./probe
[{:a 1} {:dyn true} {:tag String} {:b true, :a true} nil nil]
```

Pinned as an **ungated** `aot_e2e` case, not behind `aot_full_test`.

## 2. Accessors and predicates

Fixed in `name`, `namespace` and every type predicate. Two things beyond the report:

- The **IR interpreter had the same bug**: `ir_interp.rs`'s `KnownFn::IsSeq/IsSymbol/IsString/IsKeyword/IsNil/IsNumber/IsBool/IsInt` matched `args.first()` directly while `IsVector`/`IsMap` unwrapped. Same split, one tier down.
- Patching predicates one at a time is how the split arose, so the gate is driven by `ns-publics`, not a hand list (`tests/meta_predicate_gate.rs`): 84 predicates x 6 metadata-bearing values, 0 disagreements. It carries a positive control aimed at `meta`, which must be flagged, otherwise "0 offenders" is a claim about the gate rather than the code.

Your `(name ^{:a 1} 'x)` spelling now returns `"x"` for a different reason: `^{:a 1} 'x` is a quote form, so the annotation drops as a hint. Transparency is tested through `'^{:a 1} x` and `with-meta`, which is where wrapped symbols actually come from.

## 3-5. Should-fixes

- **`type_tag_matches`**: fixed, with the test its doc comment always implied. `tests/meta_dispatch.rs` asserts both tables agree over every value shape, bare and wrapped, with a control so "always true" can't pass.
- **`value_to_form`**: quotes the annotation now. One measured divergence worth recording: on the JVM, macro-attached metadata on a *constructing* form is compiled as code, so `^String` resolves to `java.lang.String` and `{:arglists '([x])}` genuinely fails on `x`. With no class to resolve a tag to, keeping it as the data it already was is the better analogue. Both documented in the golden record.
- **The two shorthand tables are one** `expand_meta_annotation(meta, general)`, parameterised on the only thing that differs: whether the general case is data, evaluated, or syntax-quoted. `^"Str"` and `^::kw` agree in both positions; `^42` is rejected rather than attached verbatim.

## Nits

All done. `supports_meta` gained `Queue`/`Macro`/`BoundFn`/`NativeFunction`. The eval-order asymmetry is resolved by the narrowing (a hint annotation is never evaluated, matching the JVM). All three crate READMEs document the new public API.

Your syntax-quote note was right and load-bearing: `syntax_quote.rs` now has a `FormKind::Meta` arm, so ``(let [^long y# 1] y#)`` resolves its gensym instead of failing on `y__auto__`.

## A second gap, from your first finding

`lower_quote` had no `FormKind::Meta` arm either, so `'^{:a 1} [1]` hit its catch-all and the **enclosing function declined IR lowering entirely**. Correct answers via tree-walk fallback, but silently never compiled, in exactly the macro-heavy code where quoted metadata lives. Pre-existing on `main`; only reachable once quoted metadata started carrying a value. `d7c1eb5` fixes it: inside `quote` everything is a literal, so whether the value can carry an annotation is decided statically from the form.

Worth mentioning because your suggested test, a `Tiered` variant of `quote_keeps_the_annotation_as_data`, can't be written without it. It's in `form_metadata_tiered.rs` now.

## Evidence

- `clojure-test-suite`: **5486 assertions, 0 failures**, unchanged.
- `cargo test --workspace --exclude cljrs-compiler` green. `cargo fmt` / `clippy` clean.
- Four coverage layers, because they catch different things:
  - **example**: your repros plus the JVM-measured matrix
  - **property** (`proptest`): attachment, preservation, stacking, quote/eval duality, observer-transparency as laws over generated annotations x forms
  - **golden**: 300-row behaviour matrix, both tiers, byte-compared, with the tier blocks asserted equal row for row so a regeneration can't quietly bless a divergence
  - **mutation**: 8 hand-written mutants (including reverting the IR arm to the reported bug) and `cargo-mutants` over the changed functions. 14/14 viable killed, 0 survivors.
- Non-vacuity checked, not assumed: `CLJRS_NO_IR=1` makes all 12 tier tests **fail** rather than pass by comparing two tree-walks.

The golden matrix earned its keep immediately: it caught two divergences I'd introduced myself, both cases where the tiers failed *differently* on a malformed annotation. No example test I'd have thought to write would have covered them.

crates/cljrs-ir/src/lower/anf.rs
```diff
@@ -458,7 +458,14 @@ fn lower_form(ctx: &mut LowerCtx, form: &Form) -> R {
                 && is_meta_async(meta)
             {
                 lower_fn(ctx, &parts[1..], true)
+            } else if inner.takes_runtime_meta() {
+                // The annotation becomes runtime metadata, exactly as the
+                // tree-walker's `FormKind::Meta` arm does it. Without this the
+                // same expression would answer `meta` one way interpreted and
+                // another way once promoted or AOT-compiled.
+                lower_meta_attach(ctx, meta, inner)
```
csm, line 466: **`^:async (fn …)` is the one annotation the two tiers still disagree on** — the `is_meta_async` branch just above returns `lower_fn(…, true)` and never reaches this attach.

```
$ cljrs eval '(defn g [] (meta ^:async (fn [] 1))) (println "cold:" (g)) (dotimes [_ 5000] (g)) (println "hot:" (g))'
cold: {:async true}
hot: nil
```

`takes_runtime_meta()` is true for `(fn …)`, so the tree-walker's `FormKind::Meta` arm attaches `{:async true}`; this path drops it. The arm ordering is deliberate — the cljrs-ir README records it — but the dropped annotation reads as a side effect of it rather than a decision.

By this PR's own rule the tree-walker is the consistent one: `(fn …)` constructs an `IObj`, so wrapping the `lower_fn` result the way `lower_meta_attach` does would close it. Consuming `:async` in the tree-walker too is the other coherent answer. Either is fine — right now neither tier is deferring to the other.

One adjacent thing worth a look while you're here: the tree-walker detects `:async` by peeling `^meta` off the fn's *internal* parts (`special.rs:186`), not off an outer `^:async (fn …)`, so this branch may be the only place that spelling produces an async fn at all. I couldn't make that observable through `meta` alone, and it's pre-existing either way.

Coverage: `ANNOTATIONS` in `meta_golden.rs:73` has no `:async` row (`grep -c async tests/golden/reader_metadata.txt` → 0), and it is the one annotation with a second consumer. `assert_same_in_every_tier("(meta ^:async (fn [] 1))", …)` would have caught this in a line.

---

crates/cljrs-reader/src/form.rs
```diff
@@ -34,6 +34,64 @@ impl Form {
...
+            FormKind::Vector(_) | FormKind::Map(_) | FormKind::Set(_) | FormKind::AnonFn(_) => true,
+            // A list is a call, except when it *is* a function form.
+            FormKind::List(parts) => matches!(
+                parts.first().map(|f| &f.kind),
+                Some(FormKind::Symbol(s)) if s == "fn" || s == "fn*"
+            ),
+            // Metadata stacks: `^:a ^:b [1]` annotates the vector twice.
+            FormKind::Meta(_, inner) => inner.takes_runtime_meta(),
+            _ => false,
+        }
```
csm, line 92: **`()` falls between the two rules, and the quoted and evaluated positions disagree on it.**

```
(meta '^{:a 1} ())  ;=> {:a 1}
(meta ^{:a 1} ())   ;=> nil
(meta ^{:a 1} [])   ;=> {:a 1}
```

The `List` arm classifies by head symbol, and an empty list has no head, so `parts.first()` is `None` and `()` lands in "it's a call". But it isn't a call — it's the empty-list literal, and its sibling empty literals `[]`, `{}` and `#{}` all construct an `IObj` and attach. Both tiers agree on `nil` here, so this isn't a tier bug; it's a hole in the rule's shape, and `quoted_value_supports_meta` (which says `List` → true) disagrees with it on the same form.

For what it's worth the JVM attaches here too, and for the reason that makes `()` and `[]` the same case: `Compiler.analyze` has an empty-`IPersistentCollection` branch that wraps `EmptyExpr` in a `MetaExpr`, and it runs *before* the `ISeq` dispatch that would otherwise send `()` to `analyzeSeq`. I don't have a JVM here to confirm, so worth one check before deciding which way to go — but the internal asymmetry stands either way.

Nothing flagged it because `()` is in neither generator: not in `FORMS` (`meta_golden.rs:87`), and in neither `constructing_form_strategy` nor `hint_form_strategy` (`meta_properties.rs:133`). `[]` *is* in the constructing strategy, so `[]`/`()` is exactly the pair the axes miss.

---

crates/cljrs-reader/src/form.rs
```diff
@@ -34,6 +34,64 @@ impl Form {
...
+    ///
+    /// True when the value this form denotes *as data* can carry metadata.
+    ///
+    /// Inside `quote` every form is a literal, so whether an annotation lands
+    /// is a question about the form and needs no runtime test: `'^{:a 1} [1]`
+    /// carries it, `'^{:a 1} 42` cannot. Mirrors `supports_meta` in
+    /// `cljrs-runtime` over the values `form_to_value` produces — a reader
+    /// macro (`'x`, `@x`, `#'x`, `` `x ``) denotes a list, and `#(…)` denotes
+    /// the `fn*` list it expands to.
+    pub fn quoted_value_supports_meta(&self) -> bool {
```
csm, line 59: The two doc comments got crossed. Everything from line 37 to 57 is one block attached to `quoted_value_supports_meta`, and its first three paragraphs ("True when an evaluated-position `^meta` annotation … becomes *runtime* metadata", "Only a form that constructs an `IObj` qualifies", "Every execution tier consults this one predicate") describe `takes_runtime_meta`. Line 51 is where its own doc actually starts.

`takes_runtime_meta` is then left with only line 79-80 — "Inside `quote` the rule does not apply: there the annotation is data and lands on any value that can carry it" — which is the closing sentence of the *other* function's doc and reads, on the function it now sits above, as a description of the wrong predicate.

The reader README has the correct text for `takes_runtime_meta`, so this is a slipped paste rather than a misunderstanding — but these are the two predicates the whole rule rests on, and right now each one's doc points at the other.

While you're here: `quoted_value_supports_meta` is public and new in `d7c1eb5`, and isn't in `cljrs-reader/README.md` — the README pass landed in `a6f14a1`, one commit earlier, so it's the single item that missed it.

---

csm (commented):
## Re-review — `d7c1eb5`

**Both blockers are fixed and all seven items are closed.** Two new findings, neither a repeat.

CI is green on this head; I re-ran the repros against a debug build of `d7c1eb5` (`cargo build --workspace` clean). My own full-workspace run was still building when I wrote this, so CI is the suite signal here, not me.

### Closed

| | |
|---|---|
| Tier divergence | `cold: {:a 1}` / `hot: {:a 1}`. Stacked and evaluated-map annotations agree in both tiers too. |
| Symbol accessors | `(name (with-meta 'x {:a 1}))` → `"x"`, `symbol?` → `true`, the name-mangling macro returns `1`. |
| `type_tag_matches` | Unwraps; `meta_dispatch.rs` pins the agreement rather than leaving it to the doc comment. |
| `value_to_form` | Quotes what it rebuilds — `{:tag 'String}` and `{:arglists '([x])}` both return `[1]`. |
| Shorthand drift | One `expand_meta_annotation`; `^"String"` and `^::foo` agree across positions; `^42` rejected. |
| syntax-quote | `` `(let [^long y# 1] y#) `` → `1`. |
| READMEs | All three, with signatures — one omission noted inline. |

The fix went past what I flagged, and in the right direction. Narrowing to "the form constructs an `IObj`" is Clojure's actual rule, not an approximation of it — `Compiler.analyze` attaches `MetaExpr` in `VectorExpr.parse` / `MapExpr.parse` / `SetExpr.parse` and in `FnExpr.parse`, and nowhere in `analyzeSeq` — so `(meta ^{:a 1} (list 1))` → `nil` is correct rather than a regression, and putting it in one predicate both tiers consult is the shape that keeps it correct. Lowering to `meta` → `merge` → `with-meta` through the ordinary `LoadGlobal`+`Call` path is a sound trade for something this rare.

### New

- **`^:async (fn …)` still diverges between tiers** — `{:async true}` interpreted, `nil` once promoted. The `is_meta_async` branch returns before the attach. Detail on `anf.rs:461`.
- **`()` falls between the rules** — `'^{:a 1} ()` is `{:a 1}` but `^{:a 1} ()` is `nil`, while `^{:a 1} []` is `{:a 1}`. Detail on `form.rs:81`.

Both sit *outside* the generators rather than being mis-asserted by them: `:async` isn't in `ANNOTATIONS` and `()` isn't in `FORMS` or either proptest strategy. That's the honest failure mode for a suite built this way — the axes are right, they're each missing one value.

### Nits

- The two new predicates' doc comments are crossed — detail on `form.rs:37`.
- `takes_runtime_meta` matches the head symbol textually, so `(meta ^{:a 1} (clojure.core/fn [] 1))` → `nil` where `(fn …)` and `(fn* …)` both give `{:a 1}`.
- `meta_predicate_gate.rs` pins `TreeWalk`, so its `ns-publics` guarantee — "a predicate added later is checked without anyone remembering" — doesn't reach the `KnownFn` copies in `ir_interp.rs`, which is the second place predicates live and the place the drift came from. `form_metadata_tiered.rs:176` spot-checks four in both tiers, which covers today but not the drift the gate exists to prevent.
- Roughly nineteen of the predicates given `unwrap_meta()` in `builtins.rs` are for types `supports_meta` rejects (`zero?`, `true?`, `string?`, `char?`, `var?`, `atom?` …), so the unwrap can't fire there. Harmless, and a uniform rule is defensible — just noting it's wider than the fix needed.
- `lower_meta_attach` resolves `clojure.core/meta`, `merge` and `with-meta` at run time, so redefining any of those changes what an annotation means in compiled code but not in the tree-walker. Exotic enough not to matter; the trade is well argued in the ir README and could carry a sentence there.

### On the evidence

The test layer is the strongest part of the change: positive controls on both gates (`the_gate_can_fail`, `CLJRS_NO_IR=1`), `assert_actually_lowered` so a tier comparison can't pass vacuously, and a golden matrix byte-compared across positions with the blocks asserted equal row for row. The deliberate divergences from 1.12 being written down and pinned rather than discovered later is the right call.

---

csm:
Filed #359 for the two findings in my last review — neither blocks this PR, and it's good to ship as is.

The full-workspace run I said was still building finished green: 1214 tests across 140 targets, 0 failures (`--exclude cljrs-compiler`, same scope as yours; CI covers the AOT case).

---

PR #347, opened 2026-08-20
feat(core): add eval and the host clock

BuddhiLW:
## What

- `System/currentTimeMillis`, `System/nanoTime` (monotonic, origin fixed at
  startup via `init_clock`), and `Thread/sleep`.
- `eval`, which converts the form *value* back into a `Form` and evaluates it at
  top level of the current namespace — vars visible, the caller's locals not, as
  on the JVM.

The `System/…` and `Thread/…` names follow the existing `Math/…` convention: a
JVM static's name registered as an ordinary builtin, so portable code that
reaches for a clock resolves without a reader conditional.

## A dedup that came with it

Three evaluators each kept their own list of natives that need form-level
interception, so `eval` worked under `cargo test` and threw from the CLI. That
list now lives once, in `cljrs_runtime::interp::apply::is_form_intercepted`, and
the async tree-walker and IR interpreter consult it instead of copying it.

## Evidence

`crates/cljrs-runtime/tests/eval_and_clock.rs`.

---

hierarchies change.

Stacked on #344, #345 and #346 — please merge those first.


csm:
init_clock() will panic the WASM build at startup. register_all now unconditionally calls init_clock(), which eagerly calls Instant::now(). On wasm32-unknown-unknown, Instant::now() panics unconditionally at runtime ("time not implemented on this platform") — there's no clock on that target without a JS shim. Since RuntimeBuilder::build() (runtime.rs:194) calls register_all, and the browser REPL builds a Runtime on page load, this turns a latent "calling nanotime in the browser panics" into "the whole WASM REPL traps at startup". The WASM CI job only compiles (wasm-pack build), so it will never catch this. Suggest either #[cfg(not(target_arch = "wasm32"))]-gating the eager init (the OnceLock already makes the origin lazy — the eager call only buys "origin at startup"), or using web-time/instant for a JS-backed clock.

Formatting gate is failing in CI, scrub formatting issues.

PR #348, opened 2026-08-20
feat(core): add future, cooperative on the isolate executor

BuddhiLW:
## What

`future`, `future-call`, `future?`, `future-done?`, `future-cancelled?`,
`future-cancel`. `(future body…)` expands to `(future-call (fn [] body…))`,
which hands the thunk to the installed async runtime's executor and returns a
`Value::Future` immediately.

## How this differs from the JVM, and why

Every Clojure value is `!Send` — `GcPtr` cannot cross a thread, which is what
makes per-isolate heaps collectable without cross-thread coordination. A future
therefore cannot run on another OS thread; it is a task on the *caller's*
executor. Consequences, all documented in both crate READMEs:

- The body does not start until the caller yields. `(let [f (future (f!))] …)`
  runs nothing until something awaits.
- Read a future with `(await f)`, not `@f`: `deref` blocks the one thread the
  task needs in order to finish.
- `future-cancel` marks the result cancelled (awaiting it raises) but does not
  interrupt a task already running, like a JVM future past its interrupt point.

Genuine parallelism across threads remains the isolate boundary
(`cljrs-async`'s `Isolate` + `isolate-chan`), where values cross by copy.

## Evidence

`crates/cljrs-async/tests/future_family.rs`.

---

`eval`/clock change.

Follow-up, not in this PR: `@f` in *sync* code currently hangs rather than
erroring. Fixed locally by stamping the driving thread on the future and
refusing a wait that thread could never wake; happy to send it as a separate PR
on top of this one.

Stacked on #344, #345, #346 and #347 — please merge those first. The `@f`-hangs-in-sync-code follow-up mentioned above is ready and can go up as soon as this lands.


csm:
Overall: solid design, well-documented, well-tested — but one real correctness bug in cancel semantics that I'd want fixed before merge.
What's good
 • Turns dead code into a coherent feature. The two #[allow(dead_code)] predicate stubs and the "not yet implemented" future-call placeholder are replaced by real registrations, and the old builtin_future_cancel bug (returned true even when the future had already settled) is fixed by delegating to the new CljxFuture::cancel(), which correctly returns false post-settlement.
 • The cooperative-not-parallel framing is honest and thorough. The !Send/GcPtr rationale, the "body doesn't start until you yield" semantics, the "use await, never @f" rule, and the non-interrupting future-cancel are all documented in both READMEs and in the docstrings users will actually hit ​￼.
 • Tests pin the contract deterministically — including the JVM-divergent parts (the_body_does_not_run_before_the_caller_yields would be a flaky timing assertion on the JVM; here it's exact). Type-error paths and realized? interop are covered too.
 • The PR's "stacked on #344–#347" note is now stale in a good way: those PRs are no longer open, the branch merges cleanly into main, and the diff is just these 6 files (+288/−46). CI is still running on the head commit ​￼.
Blocking: cancel is not sticky — settle_future can resurrect a cancelled future
CljxFuture::cancel() sets FutureState::Cancelled, but settle_future in crates/cljrs-async/src/eval_async.rs unconditionally overwrites the state when the task completes:
￼
```rust

pub(crate) fn settle_future(future: &GcPtr<CljxFuture>, result: EvalResult) {
    let mut state = future.get().state.lock().unwrap();
    *state = match result {
        Ok(v) => FutureState::Done(v),
        ...
```

cancel() doesn't abort the task (deliberately — "the task is not interrupted"), so this interleaving is reachable:

```clojure
(let [p (promise)
      f (future (do (await p) :v))   ; starts, parks on (await p)
      g (future 1)]
  (await g)                          ; f's task is now parked mid-body
  (future-cancel f)                  ; => true, state = Cancelled
  (deliver p nil)
  (await g)                          ; executor resumes f's task; it completes,
                                     ; settle_future overwrites Cancelled -> Done(:v)
  [(future-cancelled? f) (await f)]) ; => [false :v] — was [true ...] a moment ago
```
￼
future-cancelled? flips from true back to false, and await yields the value instead of raising. That's non-monotonic cancellation — it contradicts the PR's own docstring ("so awaiting it raises") and diverges from the JVM, where isCancelled never un-becomes true. The existing tests don't catch it because they only cancel before the first yield, where await observes Cancelled before the task ever runs.
Suggested fix (one guard in settle_future):

```rust
let mut state = future.get().state.lock().unwrap();
if matches!(&*state, FutureState::Cancelled) {
    return; // cancel wins: run the task to completion, discard the result
}
```
￼
Plus a test with the interleaving above. This keeps the documented "side effects still happen, result is discarded" contract intact.
Non-blocking nits
1. Inconsistent cancelled-error representation between the two await paths. The tree-walker (await_value) raises EvalError::Runtime("future was cancelled"); the compiled state-machine path (check_ready in state_machine.rs) reports Readiness::Failed(Value::Str("future was cancelled")) — a raw thrown string. Same condition, different error shapes; code that catches and inspects the error will behave differently depending on whether the enclosing fn got async-JIT-compiled. Worth unifying (e.g. both producing the same catchable error value).
2. builtin_future_call's args.first().cloned().unwrap_or(Value::Nil). Arity is Fixed(1), so the fallback is dead code — and if it ever fired, it would spawn a task that fails confusingly at call time (applying nil) rather than erroring at the call site. args[0].clone() or a proper arity error is clearer.
3. Dynamic bindings are not conveyed — worth a doc line. binding is a thread-local stack with an RAII guard, and future-call builds a fresh Env for the task, so (binding [*foo* 1] (future *foo*)) sees the root binding (JVM future conveys the frame via bound-fn). There's also a subtler hazard: a task polled while another task's binding frame is held across a yield could observe that frame. This is arguably a pre-existing property of the whole async substrate rather than something this PR introduces, but future makes it reachable from ordinary code — the "differs from the JVM" list is the natural place to mention it.
4. PR body housekeeping: the "stacked on #344–#347, merge those first" note can be dropped now that those have landed; CI on the head commit was still in progress when I checked, so verify it's green before merging ​￼.
Verdict
Approve after the settle_future cancel-stickiness fix (it's a two-line guard plus one test). The rest is polish. The cooperative-executor semantics are the right call for a !Send value model, and the level of honesty in the docs about where this diverges from the JVM is exactly what this kind of port needs.

PR #362, opened 2026-08-29
refactor(core): keep clojure.core/future undefined; land the cooperative one as cljrs.core.experimental/future (continues #348)

csm:
Continues #348, which could not be pushed to directly (its head lives on a fork). This branch carries @BuddhiLW's original commits unchanged, then adds two commits: one addressing the review feedback on that PR, and one moving the whole future family out of `clojure.core` after CI showed why it cannot live there.

## Why `clojure.core/future` stays undefined

`main` had already made this call deliberately — `builtins.rs` carried `// future/agent: not yet implemented — omitted from namespace`, and the standard test suite's `(when-var-exists future …)` guards are the other half of that contract. Defining the var flipped those guards on, and every one of them failed.

Of the four guarantees `clojure.core/future` carries, a one-thread-per-isolate runtime can provide one:

| | provable here? |
|---|---|
| Body runs in parallel | No — every value is `!Send`, one thread per isolate |
| `@f` blocks the caller until done | No — the caller *is* the thread the body needs |
| Binding conveyance | Yes, mechanically |
| `future-cancel` interrupts the thread | No — cooperative, no preemption |

The CI failure was four errors reading `future requires an async runtime`; locally the same tests hang, because `@f` — and `(await f)` from inside any ordinary `defn` — parks the executor thread the task needs to finish. Both were reproduced directly.

## What this PR does

The cooperative implementation is worth keeping for integrators who want it within its real limits, so it moves rather than going away:

- `builtins::register_experimental` interns `future-call`, `future?`, `future-done?`, `future-cancelled?` and `future-cancel` into `EXPERIMENTAL_NS` (`cljrs.core.experimental`). The shared intern-with-docs loop is factored out of `register_all`, and the doc strings move to their own `EXPERIMENTAL_DOCS` table so each is checked against the namespace it is actually registered in.
- The `future` macro moves from `bootstrap.cljrs` to a new `experimental.cljrs`, expanding to a fully-qualified `future-call` so it works through an alias or a bare `require`. Its docstring states up front which four guarantees it does not make.
- `Runtime::build` registers and evaluates the namespace but does not refer it into `user` — an explicit `require` is the only way in, so feature probes keep finding nothing in `clojure.core`.

```clojure
(require '[cljrs.core.experimental :as x])
(defn ^:async fetch-both [] [(await (x/future (load-a))) (await (x/future (load-b)))])
```

Three pre-existing `#[ignore]`d tests that encode the JVM contract (`test_future`, `test_binding_conveyance`, `deref_of_future_in_sync_context_still_works`) keep their ignores, with reasons that now say "by design" rather than "not yet implemented".

## Review feedback from #348 (first commit)

**Blocking — `future-cancel` did not stick.** `cancel()` does not interrupt the running task, so a future cancelled while parked mid-body kept executing and `settle_future` then overwrote `Cancelled` with the task's result: `future-cancelled?` flipped back to false and `await` yielded a value it had already raised on. `settle_future` now returns early on an already-cancelled future — the task still runs to completion and its side effects still happen, but the result is discarded. It is the only writer of a settled state in the workspace, so the guard covers every path. The regression test drives that interleaving and fails without the guard.

**Nits, all addressed:**

- **One error shape for cancellation.** There were three — `EvalError::Runtime` in the tree-walking await, a raw `Value::Str` from the compiled state machine's `check_ready`, and `ValueError::Other` in `deref`. All five sites now raise `CljxFuture::cancelled_error()`, one catchable `ex-info`-like error carrying `FUTURE_CANCELLED_MSG`, so a `catch` behaves the same whether or not the enclosing fn was async-JIT-compiled.
- **`future-call` arity.** It took `args.first().cloned().unwrap_or(Value::Nil)`; it now reports an arity error at the call site.
- **Dynamic bindings documented**, in both directions: the binding frame is not conveyed into a task, and tasks sharing the thread's binding stack can observe another task's frame across a yield.

Per CLAUDE.md, `cljrs-value/README.md` documents `CljxFuture`'s public API (`cancel`, `cancelled_error`, `FUTURE_CANCELLED_MSG`), and the `cljrs-runtime` / `cljrs-async` READMEs cover the new namespace, the new source file, and the new test file.

The "stacked on #344–#347" note in #348's description is left as-is: those have since merged, and the note is useful historical context.

## Testing

The four namespaces that failed CI now report `SKIP - future` and pass — 55 assertions, 0 failures, 0 errors, exit 0 — both AOT-compiled via `cljrs compile --test` and interpreted via `cljrs test`, where before they errored under AOT and hung under the interpreter.

| Check | Result |
|---|---|
| `cargo fmt --all --check` | clean |
| `cargo clippy --workspace -- -D warnings` | clean |
| `cljrs-runtime` + `cljrs-async` | 620 passed, 0 failed |
| `cljrs-value`, `cljrs-compiler`, `cljrs` | pass (verified before the merge from main) |

Tests were run per-crate rather than as `cargo test --workspace` — the workspace `target/` exceeded the available disk allowance — so CI is the first full-workspace run.

## Left alone

Found while diagnosing, unrelated to `future`: the AOT test harness generated by `cljrs compile --test` has no async runtime at all, so `go`, channels and the experimental futures are all unavailable inside a compiled test binary. That is what made the CI failure read `future requires an async runtime` rather than hanging. Worth its own issue.

https://claude.ai/code/session_01WGtxdiCbApwNCfUJcVogzF


csm:
The implementation here has too many differences from Clojure to have this be an official `clojure.core/future` implementation. The hanging behavior of `deref` in particular makes clojure-test-suite unrunnable. I'm keeping the implementation, but not under `clojure.core/future`. I've put it under `cljrs.core.experimental/future` for now.

I'd entertain an implementation based off of `Isolate::spawn` along with isolate channels. Sketch of the idea would be to spawn a new isolate, run the target function under that isolate, then pass the value back via an isolate channel to the promise generated by the future call. Deref/@ would need some care to implement, but on the surface I think its possible.

PR #365, opened 2026-08-29
fix(interp): defrecord fields in method bodies; param shadows the fn's own name

csm:
Two lexical-scope defects, both silent until call time.

1. A defrecord's fields were unbound as bare symbols in a protocol method body
   ((mutable? [_] (valid-sha? sha)) threw "unbound symbol: sha"), because a
   method impl is built as an ordinary fn whose only bindings are its params,
   while Clojure compiles the fields as instance fields in scope. build_impl_fn
   now wraps the body in (let* [f (:f this) ...] body) for each field a param
   does not already shadow, threaded from defrecord through register_impls_for_tag.
   reify (no fields) and extend-type/extend-protocol are unaffected.

2. A defn/fn param sharing the function's OWN name did not shadow it: the
   self-reference was bound AFTER the params in the same frame, overwriting the
   param. (defn text [text] {:text text}) returned the function as its :text.
   Binding the self-reference before the params, in both the interp and tiered
   apply paths, lets the param shadow it as Clojure does while the name stays
   visible in the body.

Verified against the JVM: bare-field access, mixed field/this, param shadowing,
assoc-then-read, empty-field records, reify, extend-type, self-recursion,
name-in-body identity, arity-specific and destructured and rest params.


csm:
Continues #353

PR #369, opened 2026-09-02
Add cljrs-num bulk numeric kernels; fix GC trace of primitive arrays

andrewdean:
Two changes, one enabling the other:

## Fix: GC trace never visited primitive-array boxes

`Value::trace` matched every primitive-array variant (`BooleanArray` through `CharArray`) to `{}` on the grounds that they hold no child `Value`s — but the boxes themselves are `GcPtr` heap objects and must be marked. A primitive array reachable **only through a traced collection** (e.g. a map of double-arrays) was never visited, so the sweep freed it while still referenced. Symptoms observed: `aget` seeing a dropped `Vec` (length 0) and `pthread_mutex_lock` EINVAL panics, both nondeterministic. Arrays held in stack locals survived via conservative root scanning, which is why this stayed hidden until long-lived arrays were stored inside collections. The fix is one hunk: visit each array pointer (leaf `Trace` impls for `Mutex<Vec<T>>` already existed).

## New crate: cljrs-num (`cljrs.num` namespace, default-on `num` feature)

Bulk numeric kernels, a deterministic per-stream RNG, and bulk CSV emission. The design premise is numpy's: array workloads get fast by dispatching once per *array* into native loops over contiguous unboxed buffers — the runtime's existing `double-array`/`long-array` values — not by making per-element interpretation faster. rustc/LLVM auto-vectorizes the loops; no Cranelift work involved.

- **RNG**: splitmix64 core, FNV-1a stream seeding, as a `NativeObject` handle. Scalar draws (`uniform!`/`normal!`/`lognormal!`/`poisson!`/`integers!`), bulk fills, and a partial-Fisher-Yates `sample-idx!`. A unit test pins the draw sequence against a pure-Clojure reference implementation.
- **Kernels**: elementwise with scalar broadcast (`add`/`sub`/`mul`/`div`, `emax`/`emin`, `clamp-min`/`clamp-max`, `exp`/`log`/`log1p`/`sqrt`/`abs`/`neg`, `round`), scans and shape helpers (`cumsum`, `sum`, `lag`, `stride`, `expand-stride`, `constant`, `to-longs`/`to-doubles`).
- **`write-csv!`**: column-oriented CSV writer with typed column specs (`[:d]`/`[:l]`/`[:date]`/`[:s]`/`[:const]`), NaN-as-NULL, escaping, append mode.

Registration mirrors cljrs-base64 (session bootstrap + extension set, `regex-full`/`small-regex`/`deps` pass-throughs). Crate README documents the full catalog with examples.

## Motivation & measurements

Built for a FIBO capital-markets test-data generator ported from Python/numpy to cljrs. Through these kernels its 454k-row OHLCV price table drops from ~9 minutes interpreted to ~0.3 s, and the market-data stage beats the numpy original (5.1 s) end to end. The GC fix was found when the generator started holding per-instrument price arrays in a map and they were swept mid-run.

## Testing

- 9 unit tests in the crate (RNG reference vector, kernel semantics, CSV/date formatting)
- `cargo test --workspace`: 161 suites, all green
- `cargo clippy -p cljrs-num` clean
- Exercised end to end by the external generator: 52-table dataset, 20 cross-table validation checks pass

https://claude.ai/code/session_01APez7FAHXPfRFtqd478TjG

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -752,6 +752,7 @@ dependencies = [
  "cljrs-lsp",
  "cljrs-net",
  "cljrs-nrepl",
+ "cljrs-num",
  "cljrs-project",
  "cljrs-reader",
  "cljrs-runtime",
@@ -961,6 +962,16 @@ dependencies = [
  "miette",
  "thiserror",
  "tokio",
+]
+
+[[package]]
+name = "cljrs-num"
+version = "0.1.0"
+dependencies = [
+ "cljrs-gc",
+ "cljrs-interop",
+ "cljrs-runtime",
+ "cljrs-value",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,7 @@ members = [
     "crates/cljrs-lsp",
     "crates/cljrs-nrepl",
     "crates/cljrs-base64",
+    "crates/cljrs-num",
     "crates/cljrs-tx",
 ]
 resolver = "2"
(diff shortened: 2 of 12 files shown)


andrewdean:
Withdrawing — keeping this on my fork for now.

csm:
I’ll open an issue for the GC bug, that’s an important fix.

csm:
Also, it is possible (or at least intended, I have not validated this) to publish cljrs-num as your own package that can be referenced via crate or via the cljrs.edn file (similar to deps.edn).

PR #371, opened 2026-09-02
Fix primitive array GC tracing

### LINKED ISSUE #370: bug: tracing GC not tracing primitive arrays correctly
Value::trace matched every primitive-array variant (BooleanArray through CharArray) to {} on the grounds that they hold no child Values — but the boxes themselves are GcPtr heap objects and must be marked. A primitive array reachable only through a traced collection (e.g. a map of double-arrays) was never visited, so the sweep freed it while still referenced. Symptoms observed: aget seeing a dropped Vec (length 0) and pthread_mutex_lock EINVAL panics, both nondeterministic. Arrays held in stack locals survived via conservative root scanning, which is why this stayed hidden until long-lived arrays were stored inside collections. The fix is one hunk: visit each array pointer (leaf Trace impls for Mutex<Vec<T>> already existed).

csm:
Summary:
- mark primitive-array GcPtr boxes from Value::trace
- add a deterministic regression test covering all eight primitive-array variants
- keep the unrelated cljrs-num changes from the withdrawn PR #369 out of this fix

The regression test confirms that, before the fix, a rooted collection survives while its eight array boxes are swept after two GC cycles. With the fix, all nine allocations survive.

Fixes #370

Testing:
- cargo test --workspace
- cargo test -p cljrs-value --no-default-features --features no-gc,regex-full --test primitive_array_gc
- cargo fmt --all -- --check
- git diff --check

### DIFF
--- a/crates/cljrs-value/tests/primitive_array_gc.rs
+++ b/crates/cljrs-value/tests/primitive_array_gc.rs
@@ -7,6 +7,7 @@ use cljrs_value::{PersistentVector, Value};
 
 #[test]
 fn primitive_arrays_reachable_through_a_collection_are_traced() {
+    let _frame = cljrs_gc::push_alloc_frame();
     let heap = GcHeap::new();
     let arrays = [
         Value::BooleanArray(heap.alloc(Mutex::new(vec![true]))),
@@ -20,11 +21,13 @@ fn primitive_arrays_reachable_through_a_collection_are_traced() {
     ];
     let root = heap.alloc(PersistentVector::from_iter(arrays));
 
-    assert_eq!(heap.count(), 9);
+    // This sibling allocation is not stored in the vector. It must be swept,
+    // proving that the collections below reached the free path even if the GC
+    // grace period changes.
+    let _unreachable = Value::DoubleArray(heap.alloc(Mutex::new(vec![7.0])));
 
-    // Unreachable allocations have one collection cycle of grace. Keeping the
-    // vector rooted across two cycles must also keep all array boxes reachable
-    // from its Values.
+    assert_eq!(heap.count(), 10);
+
     for _ in 0..2 {
         heap.collect(|visitor| visitor.visit(&root));
     }
@@ -32,6 +35,41 @@ fn primitive_arrays_reachable_through_a_collection_are_traced() {
     assert_eq!(
         heap.count(),
         9,
-        "the rooted vector and all eight primitive-array boxes must survive"
+        "the rooted vector and all eight primitive-array boxes must survive, \
+         while the unreachable array box must not"
     );
+
+    let vector = root.get();
+    assert_eq!(vector.count(), 8);
+    match (
+        vector.nth(0),
+        vector.nth(1),
+        vector.nth(2),
+        vector.nth(3),
+        vector.nth(4),
+        vector.nth(5),
+        vector.nth(6),
+        vector.nth(7),
+    ) {
+        (
+            Some(Value::BooleanArray(boolean)),
+            Some(Value::ByteArray(byte)),
+            Some(Value::ShortArray(short)),
+            Some(Value::IntArray(int)),
+            Some(Value::LongArray(long)),
+            Some(Value::FloatArray(float)),
+            Some(Value::DoubleArray(double)),
+            Some(Value::CharArray(character)),
+        ) => {
+            assert_eq!(*boolean.get().lock().unwrap(), vec![true]);
+            assert_eq!(*byte.get().lock().unwrap(), vec![1]);
+            assert_eq!(*short.get().lock().unwrap(), vec![2]);
+            assert_eq!(*int.get().lock().unwrap(), vec![3]);
+            assert_eq!(*long.get().lock().unwrap(), vec![4]);
+            assert_eq!(*float.get().lock().unwrap(), vec![5.0]);
+            assert_eq!(*double.get().lock().unwrap(), vec![6.0]);
+            assert_eq!(*character.get().lock().unwrap(), vec!['x']);
+        }
+        other => panic!("unexpected primitive-array values: {other:?}"),
+    }
 }
--- a/crates/cljrs-value/tests/primitive_array_gc.rs
+++ b/crates/cljrs-value/tests/primitive_array_gc.rs
@@ -1,0 +1,37 @@
+#![cfg(not(feature = "no-gc"))]
+
+use std::sync::Mutex;
+
+use cljrs_gc::{GcHeap, GcVisitor as _};
+use cljrs_value::{PersistentVector, Value};
+
+#[test]
+fn primitive_arrays_reachable_through_a_collection_are_traced() {
+    let heap = GcHeap::new();
+    let arrays = [
+        Value::BooleanArray(heap.alloc(Mutex::new(vec![true]))),
+        Value::ByteArray(heap.alloc(Mutex::new(vec![1]))),
+        Value::ShortArray(heap.alloc(Mutex::new(vec![2]))),
+        Value::IntArray(heap.alloc(Mutex::new(vec![3]))),
+        Value::LongArray(heap.alloc(Mutex::new(vec![4]))),
+        Value::FloatArray(heap.alloc(Mutex::new(vec![5.0]))),
+        Value::DoubleArray(heap.alloc(Mutex::new(vec![6.0]))),
+        Value::CharArray(heap.alloc(Mutex::new(vec!['x']))),
+    ];
+    let root = heap.alloc(PersistentVector::from_iter(arrays));
+
+    assert_eq!(heap.count(), 9);
+
+    // Unreachable allocations have one collection cycle of grace. Keeping the
+    // vector rooted across two cycles must also keep all array boxes reachable
+    // from its Values.
+    for _ in 0..2 {
+        heap.collect(|visitor| visitor.visit(&root));
+    }
+
+    assert_eq!(
+        heap.count(),
+        9,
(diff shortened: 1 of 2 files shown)


crates/cljrs-value/tests/primitive_array_gc.rs
```diff
@@ -0,0 +1,37 @@
...
+    // from its Values.
+    for _ in 0..2 {
+        heap.collect(|visitor| visitor.visit(&root));
+    }
+
+    assert_eq!(
+        heap.count(),
+        9,
+        "the rooted vector and all eight primitive-array boxes must survive"
+    );
```
csm: **No negative control: this test goes vacuously green if `GC_INITIAL_LIVES` ever changes.**

The `0..2` here is the minimum that reaches the free path, and it is coupled to `gc_header::GC_INITIAL_LIVES` (`crates/cljrs-gc/src/lib.rs:147`) with nothing asserting the coupling. Boxes start at `GC_INITIAL_LIVES - 1`; the sweep at `lib.rs:718-731` decrements an unmarked box to 0 on the first cycle and frees it on the second. Raise the constant and the loop no longer reaches the free path at all — every allocation survives for reasons that have nothing to do with `Value::trace`, and the assertion at line 32 passes.

That comment block at lines 140-147 explicitly presents the constant as a tuning knob ("A value of 2 gives exactly one cycle of grace"), so this is a knob someone will plausibly turn.

I verified the failure mode rather than guessing. With `GC_INITIAL_LIVES` set to 3 and the fix in this PR reverted — i.e. the bug fully present:

```
running 1 test
test primitive_arrays_reachable_through_a_collection_are_traced ... ok

test result: ok. 1 passed; 0 failed
```

Green, with the use-after-free back in the tree.

The fix is a sibling allocation the vector does not reference, asserted to *be* swept by the same loop. That pins the cycle count to observed sweep behaviour instead of to a constant read at authoring time:

```rust
    // Negative control: same shape, but never stored in the vector, so it is
    // unreachable from the root set and must be swept by the loop below.  It
    // pins the cycle count to observed sweep behaviour: if GC_INITIAL_LIVES
    // grows, this assertion fails loudly instead of the positive assertion
    // passing vacuously.
    let _unreachable = Value::DoubleArray(heap.alloc(Mutex::new(vec[7.0])));

    assert_eq!(heap.count(), 10);

    for _ in 0..2 {
        heap.collect(|visitor| visitor.visit(&root));
    }

    assert_eq!(
        heap.count(),
        9,
        "the rooted vector and all eight primitive-array boxes must survive, \
         and the unreachable array box must not"
    );
```

Re-running the same `GC_INITIAL_LIVES = 3` + bug-present experiment against that version:

```
assertion `left == right` failed: the rooted vector and all eight primitive-array
boxes must survive, and the unreachable array box must not
  left: 10
 right: 9
```

`left: 10` says nothing was swept at all, which is exactly the diagnosis you want: the test is no longer exercising the free path, go fix the loop. And with `GC_INITIAL_LIVES` back at 2 the control still discriminates correctly — fix present passes 5/5 runs; fix reverted fails with `left: 1, right: 9`.

---

csm, line 40: **`heap.count()` is the allocator's bookkeeping, not a statement about the arrays.**

The test never dereferences a single array box, so what it actually pins is "the free list length did not shrink." Two consequences:

1. It is an aggregate over nine objects. It cannot distinguish "all eight variants were traced" from "seven were traced and something else was allocated" — which matters precisely because the bug being fixed was per-variant, living in eight separate match arms.
2. Nothing here would fail under Miri or ASAN if a box were freed and then read, because nothing reads one. The use-after-free in #370 is a *use*-after-free; this test only observes the free.

Walking the vector to read the contents back closes both, and costs a few lines. It also exercises the same path the runtime uses to reach these boxes:

```rust
    // The boxes are not merely un-freed: their contents are still intact and
    // reachable by walking the vector, which is how the runtime reaches them.
    let vector = root.get();
    assert_eq!(vector.count(), 8);
    match vector.nth(6) {
        Some(Value::DoubleArray(p)) => assert_eq!(*p.get().lock().unwrap(), vec[6.0]),
        other => panic!("expected a DoubleArray at index 6, got {other:?}"),
    }
    match vector.nth(7) {
        Some(Value::CharArray(p)) => assert_eq!(*p.get().lock().unwrap(), vec['x']),
        other => panic!("expected a CharArray at index 7, got {other:?}"),
    }
```

(`PersistentVector` exposes `count()`/`nth()`, not `len()`/`get()` — `crates/cljrs-value/src/collections/vector.rs:48,65`.)

Checking two representative variants keeps it short; going all eight is also reasonable given the bug's shape. Either way this is a strengthening, not a correctness issue — the count assertion on its own does catch today's bug.

---

crates/cljrs-value/tests/primitive_array_gc.rs
```diff
@@ -0,0 +1,37 @@
...
+        Value::BooleanArray(heap.alloc(Mutex::new(vec![true]))),
+        Value::ByteArray(heap.alloc(Mutex::new(vec![1]))),
+        Value::ShortArray(heap.alloc(Mutex::new(vec![2]))),
+        Value::IntArray(heap.alloc(Mutex::new(vec![3]))),
+        Value::LongArray(heap.alloc(Mutex::new(vec![4]))),
+        Value::FloatArray(heap.alloc(Mutex::new(vec![5.0]))),
+        Value::DoubleArray(heap.alloc(Mutex::new(vec![6.0]))),
+        Value::CharArray(heap.alloc(Mutex::new(vec!['x']))),
+    ];
+    let root = heap.alloc(PersistentVector::from_iter(arrays));
```
csm, line 22: **Minor / hygiene: no `push_alloc_frame`, so these nine headers leak into the thread's `ALLOC_ROOTS` and outlive their boxes.**

`GcHeap::alloc` unconditionally calls `register_alloc` (`crates/cljrs-gc/src/lib.rs:675`), which pushes the raw `*mut GcBoxHeader` onto the thread-local `ALLOC_ROOTS`. That vec is only ever truncated by an `AllocRootGuard` drop (`lib.rs:905-913`), and this test never pushes a frame — so after the boxes are swept, `ALLOC_ROOTS` still holds dangling headers.

Inert as the file stands today: `GcHeap::collect` traces the closure plus active/retired regions, and never calls `trace_thread_alloc_roots` — that is only reached from the runtime's root closures in `crates/cljrs-runtime/src/env/gc_roots.rs`. Nothing in this test binary walks them, and libtest gives each test its own thread, so the pollution dies with the thread.

It stops being inert the moment a second test lands in this file that goes through a runtime root-tracing path, at which point `mark_header` runs over freed memory. One line at the top of the test makes it a non-issue and matches what production does at every entry point (`runtime.rs:263`, `session.rs:379`, `ir_interp.rs:234`, and others):

```rust
    let _frame = cljrs_gc::push_alloc_frame();
```

Flagging it as a convention worth establishing in the first GC test in this crate, since it will get copied.

---

csm (commented):
Review of the regression test only. The fix in `value.rs` is correct and I have nothing to say about it — the eight arms are exactly right, and it is byte-identical to the fix in the now-withdrawn #372.

The test does its job: I confirmed it fails deterministically with the fix reverted (`left: 1, right: 9` — the vector survives, all eight array boxes are swept) and passes with it, across repeated runs. Three gaps, in descending order of how much I'd want them addressed:

1. **No negative control** (line 28). The `0..2` loop is silently coupled to `GC_INITIAL_LIVES`. I verified that raising that constant to 3 makes this test pass green *with the bug fully reintroduced*. This is the one I'd actually hold the PR for.
2. **Count-only assertion** (line 32). It observes the free but never the use, so it proves less than it could about a use-after-free, and it can't attribute a failure to a particular variant.
3. **No `push_alloc_frame`** (line 10). Latent rather than live, but this is the first integration test in `cljrs-value` and the pattern will get copied.

All three are additive — nothing needs restructuring, and 1 and 2 land in the same block. I verified the suggested patches compile and behave as described (`cargo test -p cljrs-value --test primitive_array_gc` passing 5/5 with the fix, failing with it reverted, and failing loudly on the raised-constant experiment), and `cargo fmt -p cljrs-value -- --check` is clean on them.

One note outside the diff: since #372 was withdrawn, its version of the fix carried a short comment above the arms explaining *why* a childless box still needs marking. Worth folding in here — it's the kind of thing that stops the arms getting "simplified" back into a no-op catch-all later.

---

csm:
Closes #370

crates/cljrs-value/tests/primitive_array_gc.rs
csm: Addressed in 5910706. The test now allocates an unreachable sibling array and requires it to be swept while the rooted vector and its eight arrays survive. I also verified the control by temporarily raising GC_INITIAL_LIVES to 3: the test failed with left 10, right 9, as intended.

csm: Addressed in 5910706. After checking the heap count, the test walks the rooted vector and validates the concrete variant and contents of all eight primitive arrays.

crates/cljrs-value/tests/primitive_array_gc.rs
csm: Addressed in 5910706. The test now creates a push_alloc_frame guard before allocating, so its registered allocation roots are truncated when the test exits.

csm:
Addressed the full review in 5910706, including the additional explanatory comment above the primitive-array trace arms. Verification: cargo test -p cljrs-value; focused GC and no-GC test configurations; cargo clippy -p cljrs-value --tests -- -D warnings; formatting and diff checks.

PR #372, opened 2026-09-02
Fix GC trace skipping primitive array boxes (use-after-free)

andrewdean:
`Value::trace` matches every primitive-array variant (`BooleanArray` through `CharArray`) to `{}` on the grounds that they hold no child `Value`s — but the boxes themselves are `GcPtr` heap objects and must be marked. A primitive array reachable **only through a traced collection** (e.g. a map of double-arrays) is never visited, so the sweep frees it while still referenced. Observed symptoms: `aget` seeing a dropped `Vec` (length 0) and `pthread_mutex_lock` EINVAL panics, both nondeterministic (objects get `GC_INITIAL_LIVES` sweeps of grace before dying).

Arrays held in stack locals survive via conservative root scanning, which is why this stayed hidden — it needs a long-lived array stored *inside* a collection, across enough collection cycles. We hit it building a market-data generator that keeps per-instrument price series as `double-array`s in a map.

The fix is one hunk: visit each array pointer (the leaf `Trace` impls for `Mutex<Vec<T>>` already exist).

The regression test (`crates/cljrs-value/tests/gc_trace_primitive_arrays.rs`) roots a vector of four primitive arrays, drops the thread's alloc frame so the arrays aren't pinned as allocation roots, runs several collections, and asserts `total_freed` stays flat — observing the sweep itself rather than reading through the pointer, so a regression fails cleanly instead of exercising the use-after-free. Without the fix it fails with exactly the four array boxes freed; with it, it passes.

For what it's worth, this fix has been running in my fork's lineage (which fixed the same bug independently after a 2026-08-27 optimizer-suite panic) and under a heavy array workload since.

https://claude.ai/code/session_01APez7FAHXPfRFtqd478TjG

### DIFF
--- a/crates/cljrs-value/src/value.rs
+++ b/crates/cljrs-value/src/value.rs
@@ -1303,14 +1303,19 @@ impl cljrs_gc::Trace for Value {
             Value::Agent(p) => visitor.visit(p),
             Value::TypeInstance(p) => visitor.visit(p),
             Value::ObjectArray(p) => visitor.visit(p),
-            Value::BooleanArray(_)
-            | Value::ByteArray(_)
-            | Value::ShortArray(_)
-            | Value::IntArray(_)
-            | Value::LongArray(_)
-            | Value::FloatArray(_)
-            | Value::DoubleArray(_)
-            | Value::CharArray(_) => {}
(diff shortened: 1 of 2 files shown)


andrewdean:
Withdrawing for now.

PR #376, opened 2026-09-04
Fix hierarchy multimethod follow-ups

### LINKED ISSUE #360: Hierarchy dispatch follow-ups: method caching, stable ambiguity message, isa? duplication, methods key fidelity
Follow-ups identified while reviewing #346 (ad-hoc hierarchies + hierarchy dispatch). None of these block that PR — it is correct against `clojure/core.clj` and the vendored `clojure.core-test.derive` / `.underive` namespaces pass in both the interpreted (`cljrs test --src-path`) and AOT (`cljrs compile --test`) paths. These are the gaps left open.

## 1. No method cache — the miss path is the only path for hierarchy multimethods

`hierarchy_method_key` (`crates/cljrs-runtime/src/interp/apply.rs`) re-scans every entry in `dispatch_vals` on each call, running `isa_in_global_hierarchy` against each one, then applies an O(m²) domination filter over the matches — all while holding the `mf.methods` mutex.

Clojure caches the resolved method per dispatch value (`method-cache`) and invalidates the cache when the hierarchy changes. Here there is no cache, and a multimethod that dispatches *through* a hierarchy never hits the exact-key fast path, so this scan is the steady-state cost of every call, not a corner case.

Additional per-call overhead on that path:

- `isa_in_global_hierarchy` does a fresh `env.globals.lookup_var("clojure.core", "global-hierarchy")` and allocates `Keyword::simple("ancestors")` on every invocation.
- For a vector dispatch value it repeats both per element, recursively.

Suggested shape: a `HashMap<String, String>` cache on `MultiFn` from dispatch key to resolved method key, plus a generation counter on the global hierarchy (bumped by `derive` / `underive` / `prefer-method` / `remove-method` / `defmethod`) to invalidate it.

## 2. Ambiguity error message is nondeterministic

The `matches` vector in `hierarchy_method_key` is built by iterating a `HashMap`, so the "Multiple methods ... match dispatch value" error lists the conflicting methods in arbitrary order. The same program prints

```
Multiple methods in multimethod 'g' match dispatch value :user/d: :user/c and :user/b, and neither is preferred
```

on one run and the two names swapped on the next. Sorting the keys before formatting would make the message stable, diffable, and assertable in tests.

Related nit in the same message: it says "and neither is preferred" but the match set can contain more than two methods.

## 3. `isa?` is implemented twice with nothing pinning the two together

There are now two independent implementations:

- the `clojure.core/isa?` `defn` in `bootstrap.cljrs`
- `isa_in_global_hierarchy` in `crates/cljrs-runtime/src/interp/apply.rs`, used by dispatch

The native copy is a reasonable call — calling back into the interpreter from `apply_value` while holding `mf.methods` would be a re-entrancy hazard — and the crate README does document the seam. But the two can silently drift the moment either side gains a feature (class-like tags, per-multimethod `:hierarchy`, etc.).

They also read the var differently today: the Rust side uses `lookup_var(...)` + root deref, while the Clojure side resolves `global-hierarchy` through the normal path (which consults the dynamic binding stack). Harmless right now since `global-hierarchy` is not `^:dynamic`, but it is a second way for them to diverge.

A shared test that exercises both against the same hierarchy table would be cheap insurance.

## 4. `methods` returns string keys, not dispatch values

`builtin_methods` in `crates/cljrs-runtime/src/builtins/builtins.rs` keys its result map by the `pr_str` rendering of the dispatch value:

```clojure
(derive ::a ::b)
(defmulti f identity)
(defmethod f ::b [_] :p)
(keys (methods f))
;; => (":user/b")   ; a string; Clojure returns (:user/b)
```

\#346 adds the `dispatch_vals` map, which holds exactly the original `Value`s needed to fix this. Same applies to the keys of `prefers`.

## 5. Minor

- **`derive` docstring.** It says "Both must be namespaced keywords or symbols", but that is only the 2-arity rule — the 3-arity accepts unqualified tags, and `clojure.core-test.derive` depends on that (`(derive (make-hierarchy) :rect :shape)`).
- **`prefer-method` does not reject conflicting preferences.** Clojure throws when `(prefers y x)` already holds; here the second `prefer-method` is accepted and the resulting match set fails the domination check at dispatch time with the generic ambiguity error instead.
- **Six new names are visible in `clojure.core`.** `named?`, `hierarchy?`, `check-hierarchy!`, `check-tag!`, `extend-relation` are declared `defn-`, but `defn-` privacy is not enforced, so all of them (plus `global-hierarchy`) are auto-referred into every namespace — `(resolve 'named?)` from a fresh `user` ns returns `#'clojure.core/named?`. Shadowing works (covered by `core_shadows.rs`), so this is cosmetic, but `named?` is a plausible user-defined name. Either enforce `defn-` or move the helpers somewhere less exposed.

csm:
## Summary

- add generation-tagged caching for inherited and default multimethod dispatch
- invalidate cached resolutions after hierarchy, method, and preference changes
- make ambiguity errors deterministic and preserve dispatch values in methods and prefers
- reject conflicting preferences and keep hierarchy helpers private
- add tree-walk, no-GC, and AOT regression coverage

## Verification

- cargo test -p cljrs-runtime
- cargo test -p cljrs-value
- cargo test -p cljrs-runtime --features no-gc --test hierarchy_dispatch
- cargo test -p cljrs-compiler --features aot_full_test --test aot_e2e test_hierarchy_multimethod_cache_and_value_keys -- --exact
- cargo check --workspace --all-targets

Fixes #360


csm (commented):
## Review

**Ran on `89527ab`:** `cargo build --workspace` ✅ · `cargo clippy -p cljrs-value -p cljrs-runtime --all-targets` ✅ clean · `cargo test -p cljrs-runtime -p cljrs-value` — one flaky failure, see #2 · 4× full `cargo test -p cljrs-runtime`.

All five items of #360 are addressed. Two changes go beyond what the issue asked for and are worth calling out: dropping the `methods` mutex before `hierarchy_method_key` (it used to be held across the entire scan), and snapshotting the ancestors map once per miss rather than once per candidate and per vector element.

Findings below, most severe first.

---

### 1. The global generation can suppress the cache entirely, not just cause "an extra miss"

`crates/cljrs-runtime/src/env/apply.rs:350` writes the cache only if the generation is unchanged after resolution. `MULTIFN_GENERATION` is one process-global counter bumped by every `defmethod`, `derive`, `underive`, `remove-method`, `prefer-method`, and every `alter-var-root` on `clojure.core/global-hierarchy` — from any thread, for any multimethod. So a concurrent mutator doesn't cost one extra miss; it prevents the entry from ever being written, and every call re-runs the full O(n·m) scan.

Confirmed with a throwaway test (not committed): a second thread calling `bump_multifn_generation()` in a loop, 200 dispatches of an inherited method → `cached_method_count() == 1` in **0/200** iterations. The doc comment at `crates/cljrs-value/src/types.rs:158` ("an unrelated mutation may cause an extra cache miss") understates this.

The skip isn't needed for correctness either: caching under the stale generation is self-correcting, because the next reader observes the newer generation and `cached_method` clears the table before reading. Either drop the `==` guard, or bounded-retry the resolution instead of skipping. Narrowing the invalidation surface — a per-`MultiFn` generation for method/prefer-table edits alongside a separate hierarchy generation — would help independently.

### 2. `inherited_resolution_populates_and_reuses_the_method_cache` is flaky

Failed at `crates/cljrs-runtime/tests/hierarchy_dispatch.rs:190` (the second `cached_method_count()` assert) in one of four full `cargo test -p cljrs-runtime` runs; passes 45/45 in isolation. Same root cause as #1 — sibling tests in the same binary call `derive`/`defmethod` on other threads, and the counter is process-global. The assertion is on an implementation counter any other thread can reset, so it will keep firing intermittently. There's no test CI on the repo (only CodeQL), so this will surface as a local-run annoyance rather than a red check.

### 3. `isa_in_global_hierarchy` is now dead code, and both READMEs still point at it

Nothing in the workspace calls it — the only references are its own definition (`apply.rs:26`) and two README mentions. Dispatch goes through `isa_with_ancestors` now. `crates/cljrs-runtime/README.md:541` still states "Hierarchy dispatch uses: `isa_in_global_hierarchy(child, parent, env)`", which is false as of this PR. Either delete the function and document `global_hierarchy_ancestors` + `isa_with_ancestors`, or keep it as the single public entry point.

### 4. The `Var::bind` hook is the load-bearing invalidation for `derive`/`underive` — and it's undocumented

`crates/cljrs-value/src/types.rs:448` string-compares every var rebind in the process against `"clojure.core"` / `"global-hierarchy"`. This is the *only* thing that makes `derive`/`underive` invalidate the cache (both go through `alter-var-root` → `bind`). Neither README says so: cljrs-value's describes `Var::bind` as "promote-on-def: updates both slots", and cljrs-runtime's says `derive`/`underive` "bump the generation" without naming the mechanism. It also puts two `str` comparisons on every `def`/`alter-var-root` in the process. Bumping explicitly from the `derive`/`underive` builtins, or flagging the Var at intern time, would keep the coupling visible.

### 5. README public-API gap (repo rule)

CLAUDE.md requires each crate README to list all public functions. `crates/cljrs-value/README.md` gained `preference_vals`, the private `method_cache` field, and the two generation fns — but not the three new public `MultiFn` methods: `cached_method(&str, u64) -> Option<String>`, `cache_method(String, String, u64)`, `cached_method_count() -> usize`. The last exists only for tests and should say so.

### 6. Stale comment now contradicted by the fix

`crates/cljrs-stdlib/src/clojure/spec/alpha.cljrs:1600` still says of `(methods mm)`: *"its keys are Display strings in this runtime, not the original dispatch values, so it can't be used to reconstruct or validate the dispatch table."* That is exactly #360 item 4, which this PR fixes. It will mislead the next person revisiting `MultiSpec`.

### 7. The "shared semantics" tests for #360 item 3 don't pin the two `isa?` implementations together

`clojure_isa_and_multimethod_dispatch_share_hierarchy_semantics` asserts `(= (isa? [::a ::a] [::b ::b]) (f ::a))`, and the AOT test does `(= (isa? ::child ::parent) (= :inherited (classify ::child)))`. Both sides are `true` in both cases, so the equality holds trivially — it would pass under a wide range of independent breakage. To actually pin them together it needs a false case and a case where the two could plausibly diverge (transitive-only ancestor, mismatched vector arity, and so on).

Separately, the divergence #360 actually flagged — the Rust side reads the var's *root* while `clojure.core/isa?` resolves through the dynamic binding stack — is neither tested nor addressed. Harmless while `global-hierarchy` isn't `^:dynamic`, but the PR body reads as though item 3 is closed.

### 8. `prefer-method` conflict detection is non-transitive

`crates/cljrs-runtime/src/builtins/builtins.rs:7932` rejects only a directly recorded reverse preference. Clojure's `MultiFn.prefers` recurses through the parents of both values, so this is accepted here and throws on the JVM:

```clojure
(derive ::c ::c-parent)
(prefer-method f ::c-parent ::b)
(prefer-method f ::b ::c)   ;; Clojure: IllegalStateException
```

`dominates` at `apply.rs:94` has the same non-transitivity (pre-existing). Fine to defer, but since this is presented as closing #360's `prefer-method` bullet, the gap belongs in the README.

### 9. `defn-` privacy is refer-only, and asymmetric

`refer_all` / `refer_core_impl` skip private vars, but `refer_named` (`env.rs:656`, the `:refer [x]` path) does not, and qualified resolution goes straight through `lookup_var` → interns with no check. So `(require '[foo :refer [secret]])` and `foo/secret` both still resolve where Clojure throws "not public". That's a defensible stopping point for what #360 called a cosmetic problem, but `:refer :all` silently skipping while `:refer [name]` silently allows deserves at least a comment in `refer_named`.

### 10. Ambiguity message lists dominated candidates

`apply.rs:106` formats `matches`, not `best`. With methods on `::b`, `::c` (where `::c` isa `::b`) and an unrelated `::x`, the real conflict is `::c` vs `::x`, but the message names all three; the `"neither"`/`"none"` choice keys off `matches.len()`, so it can say "none" when the conflict is a pair. Clojure names the conflicting pair. `"a and b and c"` also reads awkwardly — commas with a final "and" would be better. Cosmetic, but message quality is the point of this change.

### 11. Merge conflict

The branch is behind `main` and conflicts in `crates/cljrs-value/src/lib.rs` — a trivial re-export-list collision with `FUTURE_CANCELLED_MSG` added on main.

---

**Bottom line:** #1 is the one I'd fix before merging — as written, the cache is defeated by exactly the workloads (concurrent, REPL, incremental namespace loading) where it matters, and #2 is that defect leaking into the test suite. #3–#6 are documentation and dead-code cleanup the repo's own rules call for. #7–#10 are scope and polish notes.

---

PR #377, opened 2026-09-07
fix(core): make `for` a real comprehension; widen `mapv`

BuddhiLW:
Two `clojure.core` gaps with the same shape: the docstring describes the full behaviour while only the simplest case is implemented, so reading the source tells you it works. Both turned up while writing a library against the runtime rather than reading it.

## `for` read only its first binding pair

```clojure
(for [x (range 2) y (range 2)] [x y])
;; before: unbound symbol: y
```

Everything after the first `binding coll` pair was dropped, and no `:let` / `:when` / `:while` modifier was supported. `doseq`, directly above it in `bootstrap.cljrs`, had handled multiple bindings and all three modifiers from the start and documents the same binding grammar — which is what made the gap easy to miss.

It now expands to nested `mapcat`, mirroring `doseq`'s `letfn` emitter:

```clojure
(for [x (range 3) y (range 2)] [x y])
;=> ([0 0] [0 1] [1 0] [1 1] [2 0] [2 1])

(for [x (range 10) :when (even? x) :let [y (* x x)]] y)
;=> (0 4 16 36 64)
```

Rightmost binding varies fastest, and laziness is preserved through the nesting — taking three elements of a comprehension over a million returns.

### One deliberate limitation

`:while` has to *stop* its loop rather than skip an element, so it compiles into a `take-while` over that binding's own collection. That works while it still stands next to the collection, optionally behind `:let` modifiers, whose bindings are folded into the predicate:

```clojure
(for [x (range 10) :let [y (* x x)] :while (< y 9)] x)   ;=> (0 1 2)
```

A `:while` placed after a `:when` in the same binding group would have to stop a loop it can no longer see. The macro **refuses at macroexpansion time** rather than silently behaving like a `:when`; lifting that needs a loop-based expansion like Clojure's chunked one. Noted in `TODO.md`.

## `mapv` defined only its two-argument arity

```clojure
(mapv + [1 2] [10 20])
;; before: arity error
```

…despite a docstring describing "the set of first items of each coll, followed by applying f to the set of second items, until any one coll is exhausted". Each arity is now a `vec` over the matching `map` arity, which already had them.

`KnownFn::Mapv` stays pinned at arity 2 exactly as `KnownFn::Map` does — `map` is already variadic at the Clojure level with the same pinning, so off-arity calls take the fallback path that already exists.

## Tests

26 new tests in `crates/cljrs-runtime/tests/`:

- `for_comprehension.rs` (18) — iteration order, three-deep nesting, an inner binding reading an outer one, destructuring, each modifier, `:when` skipping vs `:while` stopping, laziness through nesting, the rejected `:while` placement, and that a nil body yields nils rather than vanishing.
- `mapv_arities.rs` (8) — 1 to 5 collections, shortest-collection truncation, that the result is a vector rather than a seq, and that `mapv` agrees with `map` over the same arguments.

`cargo test -p cljrs-runtime -p cljrs-stdlib` is clean.

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -85,12 +85,15 @@ Implementation roadmap for a Rust-hosted Clojure dialect. Native file extension 
 - [x] Collection ops: `conj`, `assoc`, `dissoc`, `get`, `get-in`, `assoc-in`, `update`, `update-in`, `merge`, `into`, `empty`
 - [x] Seq ops: `first`, `rest`, `next`, `cons`, `seq`, `count`, `nth`, `last`, `butlast`, `reverse`, `concat`
 - [x] Higher-order: `map`, `filter`, `reduce`, `keep`, `remove`, `mapcat`, `take`, `drop`, `take-while`, `drop-while`, `partition`, `partition-all`, `group-by`, `sort`, `sort-by`
+  - [x] `mapv` over more than one collection. Only its two-argument arity existed, so `(mapv f xs ys)` was an arity error despite a docstring describing the multi-collection behaviour. Tests: `crates/cljrs-runtime/tests/mapv_arities.rs`
 - [x] Lazy sequences: `lazy-seq`, `range`, `repeat`, `iterate`, `cycle`, `repeatedly` (via `Thunk`/`LazySeq` + `Value::Cons`)
 - [x] String functions: `join`, `split`, `trim`, `upper-case`, `lower-case`, `replace`, `starts-with?`, `ends-with?`
 - [x] I/O: `print`, `println`, `prn`, `pr`, `pr-str`, `str`, `read-string`, `slurp`, `spit`
 - [x] Math: `Math/abs`, `Math/pow`, `Math/sqrt`, `Math/floor`, `Math/ceil`, `Math/round`, `Math/log`, `Math/log10`, `Math/exp`, `Math/sin`, `Math/cos`, `Math/tan`, `Math/asin`, `Math/acos`, `Math/atan`, `Math/atan2`, `Math/sinh`, `Math/cosh`, `Math/tanh`, `Math/hypot`, `Math/PI`, `Math/E`
 - [x] Miscellaneous: `apply`, `comp`, `partial`, `juxt`, `memoize`, `constantly`, `identity`, `not`, `complement`, `gensym`, `type`, `class`, `hash`
 - [x] Core macros: `when`, `when-not`, `if-let`, `when-let`, `if-not`, `cond`, `condp`, `case`, `and`, `or`, `->`, `->>`, `as->`, `doto`, `dotimes`, `doseq`, `for`
+  - [x] `for` as a real list comprehension: multiple binding pairs (rightmost fastest) and the `:let`/`:when`/`:while` modifiers, matching the grammar `doseq` beside it already implemented. It previously read only the **first** binding pair and dropped the rest, so `(for [x xs y ys] ...)` failed with "unbound symbol: y" and no modifier was supported. Tests: `crates/cljrs-runtime/tests/for_comprehension.rs`
+  - [ ] `for`'s `:while` compiles into a `take-while` over its own collection, so it has to follow its binding directly (any number of `:let` modifiers may sit in between). A `:while` after a `:when` in the same binding group is rejected at macroexpansion time rather than silently downgraded to a `:when`; supporting it needs a loop-based expansion like Clojure's chunked one
 - [x] Namespace ops: `in-ns`, `alias`, `refer` (basic); `ns` with `:require`/`:refer-clojure` (`:exclude`/`:only`/`:rename`)
 - [x] `require` — file-based namespace loading with `:as` alias and `:refer [...]`/`:refer :all`
 - [x] `load-file` — evaluate a source file by absolute path
--- a/TODO.md
+++ b/TODO.md
@@ -85,15 +85,15 @@ Implementation roadmap for a Rust-hosted Clojure dialect. Native file extension 
 - [x] Collection ops: `conj`, `assoc`, `dissoc`, `get`, `get-in`, `assoc-in`, `update`, `update-in`, `merge`, `into`, `empty`
 - [x] Seq ops: `first`, `rest`, `next`, `cons`, `seq`, `count`, `nth`, `last`, `butlast`, `reverse`, `concat`
 - [x] Higher-order: `map`, `filter`, `reduce`, `keep`, `remove`, `mapcat`, `take`, `drop`, `take-while`, `drop-while`, `partition`, `partition-all`, `group-by`, `sort`, `sort-by`
-  - [x] `mapv` over more than one collection. Only its two-argument arity existed, so `(mapv f xs ys)` was an arity error despite a docstring describing the multi-collection behaviour. Tests: `crates/cljrs-runtime/tests/mapv_arities.rs`
+  - [x] `mapv` over one or more collections (arities 1-4 and variadic)
 - [x] Lazy sequences: `lazy-seq`, `range`, `repeat`, `iterate`, `cycle`, `repeatedly` (via `Thunk`/`LazySeq` + `Value::Cons`)
 - [x] String functions: `join`, `split`, `trim`, `upper-case`, `lower-case`, `replace`, `starts-with?`, `ends-with?`
 - [x] I/O: `print`, `println`, `prn`, `pr`, `pr-str`, `str`, `read-string`, `slurp`, `spit`
 - [x] Math: `Math/abs`, `Math/pow`, `Math/sqrt`, `Math/floor`, `Math/ceil`, `Math/round`, `Math/log`, `Math/log10`, `Math/exp`, `Math/sin`, `Math/cos`, `Math/tan`, `Math/asin`, `Math/acos`, `Math/atan`, `Math/atan2`, `Math/sinh`, `Math/cosh`, `Math/tanh`, `Math/hypot`, `Math/PI`, `Math/E`
 - [x] Miscellaneous: `apply`, `comp`, `partial`, `juxt`, `memoize`, `constantly`, `identity`, `not`, `complement`, `gensym`, `type`, `class`, `hash`
 - [x] Core macros: `when`, `when-not`, `if-let`, `when-let`, `if-not`, `cond`, `condp`, `case`, `and`, `or`, `->`, `->>`, `as->`, `doto`, `dotimes`, `doseq`, `for`
-  - [x] `for` as a real list comprehension: multiple binding pairs (rightmost fastest) and the `:let`/`:when`/`:while` modifiers, matching the grammar `doseq` beside it already implemented. It previously read only the **first** binding pair and dropped the rest, so `(for [x xs y ys] ...)` failed with "unbound symbol: y" and no modifier was supported. Tests: `crates/cljrs-runtime/tests/for_comprehension.rs`
-  - [ ] `for`'s `:while` compiles into a `take-while` over its own collection, so it has to follow its binding directly (any number of `:let` modifiers may sit in between). A `:while` after a `:when` in the same binding group is rejected at macroexpansion time rather than silently downgraded to a `:when`; supporting it needs a loop-based expansion like Clojure's chunked one
+  - [x] `for` as a real list comprehension: multiple binding pairs (rightmost fastest) and the `:let`/`:when`/`:while` modifiers
+  - [ ] Loop-based `for` expansion, so a `:while` may follow a `:when` in the same binding group
 - [x] Namespace ops: `in-ns`, `alias`, `refer` (basic); `ns` with `:require`/`:refer-clojure` (`:exclude`/`:only`/`:rename`)
 - [x] `require` — file-based namespace loading with `:as` alias and `:refer [...]`/`:refer :all`
 - [x] `load-file` — evaluate a source file by absolute path
--- a/crates/cljrs-runtime/src/builtins/bootstrap.cljrs
+++ b/crates/cljrs-runtime/src/builtins/bootstrap.cljrs
@@ -378,11 +378,7 @@
 (defn mapv
   "Returns a vector consisting of the result of applying f to the set of
   first items of each coll, followed by applying f to the set of second
-  items, until any one coll is exhausted.
-
-  The docstring always described the multi-collection behaviour; only the
-  two-argument arity existed, so `(mapv f xs ys)` was an arity error. `map`
-  already carries every arity, so each one here is a `vec` over it."
+  items, until any one coll is exhausted."
   ([f coll] (vec (map f coll)))
   ([f c1 c2] (vec (map f c1 c2)))
   ([f c1 c2 c3] (vec (map f c1 c2 c3)))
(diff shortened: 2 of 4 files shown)


crates/cljrs-runtime/src/builtins/bootstrap.cljrs
```diff
@@ -601,13 +608,71 @@
...
+                              rest-binds (next (next more))
+                              rest-binds (if (seq lets)
+                                           (cons :let (cons lets rest-binds))
+                                           rest-binds)]
+                          (list 'mapcat
+                                (list 'fn (vector x) (emit rest-binds))
+                                (list 'take-while (list 'fn (vector x) test) coll)))
+
+                        :else
+                        (list 'mapcat (list 'fn (vector x) (emit after)) coll))))))
```
csm: The single-binding case regresses ~2x because it now goes through `mapcat` plus a one-element `list` per element, where it used to be a plain `map`.

Measured on this branch (debug build, `ExecutionMode::TreeWalk`, `(count (for [x (range n)] x))`):

| n | this branch | `map` equivalent |
|---|---|---|
| 1000 | 2.27s | 1.32s |
| 5000 | 13.08s | 8.81s |
| 20000 | 89.0s | 51.0s |

Beyond the per-element allocation, `map` is `KnownFn::Map` with a pinned arity-2 fast path (`cljrs-ir/src/lower/anf.rs:3105`, `codegen.rs` → `rt_map`). `KnownFn::Mapcat` routes to `rt_mapcat`, whose only native fast path is map-over-vector (`rt_abi.rs:3603`); everything else re-enters the bootstrap Clojure `mapcat`. So the tiered and AOT paths lose the optimization too, not just the tree-walker.

The innermost binding never needs `mapcat` — it produces exactly one element per element, which is what `map` does. Emitting `map` when nothing follows the binding recovers it:

```clojure
                        :else
                        (if (seq after)
                          (list 'mapcat (list 'fn (vector x) (emit after)) coll)
                          ;; Innermost binding, no modifiers: one element per
                          ;; element, which is exactly `map`.
                          (list 'map (list 'fn (vector x) (cons 'do body)) coll))
```

I applied that locally: all 26 new tests still pass, and `(count (for [x (range 5000)] x))` drops from 13.08s to 6.25s (n=1000: 2.27s → 1.17s). The same shortcut applies to the `:while` branch above when `rest-binds` is empty.

---

crates/cljrs-runtime/src/builtins/bootstrap.cljrs
```diff
@@ -601,13 +608,71 @@
...
+              ;; Innermost: one element per surviving binding combination.
+              (list 'list (cons 'do body))))]
```
csm: This emits an unqualified `list` **inside** every generated `fn`, so a binding named `list` shadows it and the expansion calls the bound value:

```clojure
(for [list [1 2]] list)
;; not callable: <long> is not callable
```

That is a regression: the old expansion was `(map (fn [x] (do body)) coll)`, which never mentions `list`, so this worked before the PR. It also bites `:let`:

```clojure
(for [x [1 2] :let [list x]] list)   ;; same error
```

`mapcat` and `take-while` (lines 667/669/672) have the same exposure one level in, where an outer binding shadows them for a nested one:

```clojure
(for [mapcat [[1] [2]] x mapcat] x)  ;; not callable
```

`doseq` next door shares the class for `next` (`(recur (next gs))` sits inside the `let` that binds `x`), so it is a pre-existing pattern — but `list` is a much more likely binding name than `next`, and this PR is what introduces it.

Qualifying the emitted symbols fixes all of these and resolves correctly through a shadowing local (verified: `(let [list 1] (clojure.core/list 1 2))` → `(1 2)`):

```clojure
(list 'clojure.core/list (cons 'do body))
;; likewise 'clojure.core/mapcat, 'clojure.core/take-while
```

---

crates/cljrs-runtime/src/builtins/bootstrap.cljrs
```diff
@@ -601,13 +608,71 @@
...
+                        (= :while (first more))
+                        (let [test (second more)
+                              test (if (seq lets) (list 'let lets test) test)
+                              rest-binds (next (next more))
+                              rest-binds (if (seq lets)
+                                           (cons :let (cons lets rest-binds))
+                                           rest-binds)]
+                          (list 'mapcat
+                                (list 'fn (vector x) (emit rest-binds))
+                                (list 'take-while (list 'fn (vector x) test) coll)))
```
csm: Folding the `:let` bindings into the `take-while` predicate **and** re-emitting them into `rest-binds` means each `:let` init runs twice per element — once in the predicate, once in the body:

```clojure
(let [n (atom 0)]
  [(vec (for [x (range 4) :let [y (do (swap! n inc) x)] :while (< y 3)] y))
   @n])
;=> [[0 1 2] 7]     ;; Clojure: [[0 1 2] 4]
```

Four evaluations in the `take-while`, three more in the body. Clojure evaluates a `:let` init once per iteration, so anything side-effecting or non-deterministic (`swap!`, `rand`, `gensym`, an I/O read) observably differs, and an expensive init silently doubles in cost.

Neither the docstring nor the `TODO.md` note mentions this — they only cover the placement restriction. At minimum it belongs in both; better would be to bind once outside the predicate, e.g. `take-while` over a `map`ped `[x lets...]` pair, or the loop-based expansion already noted as the eventual fix.

---

crates/cljrs-runtime/src/builtins/bootstrap.cljrs
```diff
@@ -601,13 +608,71 @@
...
+                  (= :while tag)
+                  (throw (ex-info
+                          (str "for: :while must follow its binding directly "
+                               "(only :let may come between)")
+                          {:modifier :while :bindings bindings}))
+
+                  :else
+                  (let [x tag
+                        coll (second binds)
+                        after (next (next binds))]
```
csm: A malformed binding vector silently yields an empty seq instead of erroring, because `(second binds)` is `nil` and `mapcat` over `nil` is empty:

```clojure
(for [x [1 2] y])        ;=> ()   ;; Clojure: "for requires an even number of forms"
(for [x [1 2] :when])    ;=> ()
```

Now that the macro parses a real grammar this is worth a check — a dropped collection expression is an easy typo, and "returns nothing" is the worst way to report it. The `:while` branch already throws via `ex-info`, so the machinery is there:

```clojure
(when-not (next binds)
  (throw (ex-info "for requires an even number of forms in binding vector"
                  {:bindings bindings})))
```

---

crates/cljrs-runtime/src/builtins/bootstrap.cljrs
```diff
@@ -378,8 +378,15 @@
 (defn mapv
   "Returns a vector consisting of the result of applying f to the set of
   first items of each coll, followed by applying f to the set of second
-  items, until any one coll is exhausted."
-  [f coll] (vec (map f coll)))
+  items, until any one coll is exhausted.
+
+  The docstring always described the multi-collection behaviour; only the
+  two-argument arity existed, so `(mapv f xs ys)` was an arity error. `map`
+  already carries every arity, so each one here is a `vec` over it."
```
csm: This paragraph is PR-description material in a user-facing docstring — `doc` will print "only the two-argument arity existed, so `(mapv f xs ys)` was an arity error" to someone asking what `mapv` does today. The first paragraph already describes the behaviour completely. The history belongs in the commit message and the `TODO.md` entry, both of which already carry it.

Same applies to the last paragraph of `for`'s docstring: the `:while` placement restriction is genuinely worth documenting, but "rather than being silently downgraded to a `:when`" describes the alternative that was rejected, not the rule the caller has to follow.

The arities themselves match `clojure.core/mapv` exactly, and I confirmed the 4-fixed/4-variadic pair dispatches correctly here (`(mapv + [1] [2] [3] [4])` → `[10]`).

---

TODO.md
```diff
@@ -85,12 +85,15 @@ Implementation roadmap for a Rust-hosted Clojure dialect. Native file extension
...
 - [x] Miscellaneous: `apply`, `comp`, `partial`, `juxt`, `memoize`, `constantly`, `identity`, `not`, `complement`, `gensym`, `type`, `class`, `hash`
 - [x] Core macros: `when`, `when-not`, `if-let`, `when-let`, `if-not`, `cond`, `condp`, `case`, `and`, `or`, `->`, `->>`, `as->`, `doto`, `dotimes`, `doseq`, `for`
+  - [x] `for` as a real list comprehension: multiple binding pairs (rightmost fastest) and the `:let`/`:when`/`:while` modifiers, matching the grammar `doseq` beside it already implemented. It previously read only the **first** binding pair and dropped the rest, so `(for [x xs y ys] ...)` failed with "unbound symbol: y" and no modifier was supported. Tests: `crates/cljrs-runtime/tests/for_comprehension.rs`
+  - [ ] `for`'s `:while` compiles into a `take-while` over its own collection, so it has to follow its binding directly (any number of `:let` modifiers may sit in between). A `:while` after a `:when` in the same binding group is rejected at macroexpansion time rather than silently downgraded to a `:when`; supporting it needs a loop-based expansion like Clojure's chunked one
```
csm: Every other line in this file is a terse one-liner naming what exists; these three sub-bullets are multi-sentence retrospectives about what used to be broken and why it was missed. A roadmap reader six months from now wants "what works", not "what was wrong before #377" — that is the commit message's job.

The `- [ ]` entry is also not a roadmap item: it documents a shipped limitation rather than planned work, so it will sit unchecked indefinitely under a `[x]` parent. If the loop-based expansion is actually intended, phrase it as the work item (`- [ ] Loop-based `for` expansion so `:while` can follow `:when``); if not, the limitation belongs only in the docstring.

---

csm (changes requested):
Both gaps are real and the expansion strategy is sound. I checked out the branch, ran the 26 new tests (all pass), and probed the cases they don't cover. Four issues, one of them a regression against the old `for`.

## Verified correct

- Iteration order (rightmost fastest), three-deep nesting, inner binding reading an outer one, destructuring in both binding forms and `:let`.
- Laziness holds through nesting: `(take 1 (for [x (range 100) y (range 3)] (do (swap! n inc) [x y])))` evaluates the body exactly once, and the source collection is realized incrementally (2 elements for `(take 2 ...)`).
- `:while` over an infinite `(range)` terminates; `:while` before further bindings works; `:when` after `:while` works; `:while` stops only its own loop.
- `mapv` arities 1–5, shortest-collection truncation, and the 4-fixed/4-variadic pair dispatching correctly. `KnownFn::Mapv`'s arity-2 pin is not disturbed — off-arity calls take the existing fallback, as the description says.

## Issues

1. **Perf regression on the common case** (`bootstrap.cljrs:671`). `(for [x xs] body)` used to expand to `map`; it now goes through `mapcat` plus a per-element `list`. Measured ~2x slower, and it loses `KnownFn::Map`'s pinned fast path in the IR interpreter and codegen. Emitting `map` for the innermost binding when no modifiers follow recovers it — I applied that locally, tests stay green, `n=5000` drops from 13.1s to 6.3s.

2. **Hygiene regression** (`bootstrap.cljrs:674`). The unqualified `list` is emitted inside every generated `fn`, so `(for [list [1 2]] list)` now fails with "not callable". It worked before this PR. `mapcat`/`take-while` have the same exposure one nesting level in. Qualifying to `clojure.core/list` etc. fixes it; I verified qualified symbols resolve through a shadowing local.

3. **`:let` runs twice per element under `:while`** (`bootstrap.cljrs:661`). The bindings are folded into the `take-while` predicate *and* re-emitted into the body, so a side-effecting or expensive init is evaluated twice — 7 evaluations where Clojure does 4, in a 4-element loop. Not mentioned in the docstring or the `TODO.md` note, which only cover the placement restriction.

4. **Malformed binding vectors silently yield `()`** (`bootstrap.cljrs:649`). `(for [x [1 2] y])` and `(for [x [1 2] :when])` both return an empty seq rather than erroring. Clojure throws.

Minor: retrospective prose in the `mapv` docstring (it reaches `doc` output), and `TODO.md` sub-bullets written as changelog entries in a file of one-liners. Both noted inline.

One behavioural note that is **not** a defect in this PR: `(for [x [] ] x)` prints as `nil` rather than `()`, but so does `(map inc [])` on `main` — it is the runtime's existing empty-seq representation, not something the new expansion introduces.

The two-`:while`-on-one-binding case (`[x c :while p :while q]`), which Clojure accepts, is rejected here with "must follow its binding directly". That is defensible under the documented limitation, but the message doesn't describe what's actually wrong.

---

BuddhiLW:
Thanks — all four are real, and three were regressions the old one-binding-pair stub couldn't have had. Fixed in 2b8f645.

**Hygiene.** Every emitted core symbol is qualified now. Your three reproductions pass, plus `map`, `take-while` and `some?` as binding names.

**`:let` under `:while`.** Took the "bind once" route rather than documenting it. The collection is mapped to one frame per element with the `:let` inits inside the frame, so the test and the body read the same scope:

```clojure
(mapcat first
        (take-while some?
                    (map (fn [x] (let <lets> (if <test> [<contribution>] nil))) coll)))
```

A frame is `[contribution]` while the test holds and `nil` once it fails. That distinction is what keeps `:while` able to stop the loop while an inner `:when` — whose contribution is an empty seq, not nil — only skips. Your reproduction now gives `[[0 1 2] 4]`, matching Clojure, and I added a test for the `:while`-stops-vs-`:when`-skips interaction since the frame representation is what carries it.

**`map` for the simple case.** Applied. I couldn't assert the expansion shape in a test — this runtime has no `macroexpand`/`macroexpand-1` — so the test pins the two properties the substitution has to preserve instead: one element per element, and the body staying lazy under `take`.

**Malformed binding vectors.** Now rejected. I put the check on every tag rather than just bindings, so `(for [x [1 2] :when] x)` and `(for [x] x)` are caught by the same rule.

Docstring retrospectives dropped from both `mapv` and `for`; TODO entries rewritten as one-liners. The `- [ ]` is now phrased as the work item you suggested — the loop-based expansion that would lift the `:while` placement restriction — rather than a description of the shipped limitation.

On the two-`:while` case: the message now names both causes ("a `:while` must follow its binding directly (only `:let` may come between) and a binding takes at most one `:while`").

Not addressed, deliberately: the loop-based expansion itself. It would subsume the remaining restriction and is the better long-term shape, but it's a rewrite rather than a fix and belongs in its own change.

crates/cljrs-runtime/src/builtins/bootstrap.cljrs
csm: Three of the four shapes now throw. One is still silent — a trailing `:while` with no test form:

```clojure
(for [x [1 2] :while])          ;=> ()      ;; expected: even-number-of-forms error
(for [x [1 2] :let [y 1] :while])  ;=> ()   ;; same
```

The even-number check at the top of `emit` only fires when `emit` is re-entered with a one-element list. A `:while` is consumed inside the scanning `loop` instead, and that branch takes `(second more)` as the test without checking it exists — so the test expands to the literal `nil`, the predicate is never true, and the comprehension yields nothing.

Same fix shape as the one you already applied, in the `(= :while (first more))` branch:

```clojure
(when-not (next more)
  (throw (ex-info "for requires an even number of forms in the binding vector"
                  {:bindings bindings})))
```

Not blocking — I've approved. Worth folding into `a_malformed_binding_vector_is_rejected`, which covers the other three shapes.

---

csm (approved):
Re-checked `2b8f645` on a fresh checkout. Five of six findings are fixed and verified; one sub-case of the malformed-bindings check is still silent, noted on that thread and not blocking.

## Verified fixed

**Perf regression.** The innermost binding now emits `map`. Re-measured (debug build, `TreeWalk`, `(count (for [x (range n)] x))` vs the hand-written `map`):

| n | `for` | `map` |
|---|---|---|
| 1000 | 1.13s | 1.07s |
| 5000 | 8.13s | 9.24s |
| 20000 | 38.2s | 37.5s |

Parity, down from ~2x. Previously 89.0s at n=20000.

**Hygiene.** All emitted core symbols are qualified. Every shape I could construct now works, including through a `:while`:

```clojure
(for [list [1 2]] list)                                  ;=> (1 2)
(for [list [[1] [2]] x list] x)                          ;=> (1 2)
(for [mapcat [[1] [2]] x mapcat] x)                      ;=> (1 2)
(for [map [[1] [2]] x map] x)                            ;=> (1 2)
(for [first [[1] [2]] x first] x)                        ;=> (1 2)
(for [take-while [1 2 3] :while (< take-while 3)] take-while) ;=> (1 2)
(for [some? [1 2 3] :while (< some? 3)] some?)           ;=> (1 2)
(for [x [1 2] :let [list x]] list)                       ;=> (1 2)
```

The unqualified `let`/`fn`/`if`/`do` are fine — bindings named after them still work, since the special form wins in head position.

**`:let` under `:while`.** Once per element now: the counter case gives `[[0 1 2] 4]`, matching Clojure, down from 7.

The frame rework is the part I looked at hardest, since `mapcat first` over `take-while some?` has to keep "produced nothing" distinguishable from "stop". It holds — `[nil]` is a live frame, `nil` is the terminator:

```clojure
(for [x (range 6) :while (< x 4) :when (odd? x)] x)      ;=> (1 3)
(for [x (range 6) :while (< x 4) :when (even? x)] x)     ;=> (0 2)   ; first element skipped, loop continues
(for [x (range 4) :while (< x 3) :when false] x)         ;=> ()      ; every element skipped, still terminates
(for [x (range 3) :while (< x 2)] nil)                   ;=> (nil nil)  ; a nil body is not a terminator
(for [x (range 3) :let [y nil] :while (< x 2)] y)        ;=> (nil nil)  ; nor a nil :let init
(for [x (range 10) :while (< x 3) y (range 3) :when (odd? y)] [x y]) ;=> ([0 1] [1 1] [2 1])
(for [x (range 10) :let [y (inc x)] :let [z (* y 2)] :while (< z 8)] [x z]) ;=> ([0 2] [1 4] [2 6])
(for [[a b] [[1 2] [3 4] [5 6]] :while (< a 5)] (+ a b)) ;=> (3 7)
```

Laziness survives the extra layer: `(take 2 (for [x (range 1000000) :while true] x))` returns, and `(take 1 ...)` over a nested comprehension still evaluates the body exactly once.

**Docstrings and `TODO.md`.** Retrospective prose gone; the `- [ ]` entry now names the loop-based expansion as the work item. The `:while` error message covers the two-`:while` case it rejects.

## Not fixed

A trailing `:while` with no test still yields `()` rather than throwing — `(for [x [1 2] :while])`. The check added at the top of `emit` doesn't reach the `:while` branch, which consumes its test inside the scanning loop. One-line fix, on the thread.

## Cost note, not a request

The `:while` path went from two lazy layers to three (`map` → `take-while` → `mapcat`) plus a per-element vector — about 4x the plain path at n=20000 (148s vs 38.2s). That is the price of the once-per-element `:let`, and `:while` is the rare modifier, so it looks like the right trade. Worth knowing if the loop-based expansion ever gets written.

CI is green, and `cargo test -p cljrs-runtime -p cljrs-stdlib` passes locally with no failures.

---

PR #378, opened 2026-09-07
fix(interp): resolve a cross-namespace multimethod in `defmethod`

BuddhiLW:
`defmethod` could not extend a multimethod that lives in another namespace.

```clojure
(ns shapes.square (:require [shapes.core :as core]))
(defmethod core/area :square [s] (* (:side s) (:side s)))
;; defmethod: core/area is not a multimethod
```

`eval_defmethod` resolved its name with `lookup_in_ns(&env.current_ns, name)` and nothing else, so `core/area` — or `shapes.core/area` — was looked up verbatim in the *current* namespace and never found.

That forbids exactly the thing an open dispatch is for: the namespace owning the `defmulti` doesn't know its extenders, so the extenders have to name it. The message was misleading too — `shapes.core/area` is a perfectly good multimethod.

## Fix

The name now resolves the way `eval_symbol` and `binding` resolve one:

- `alias/name` through the current namespace's `:require :as` aliases,
- `fully.qualified.ns/name` directly,
- a bare name in the current namespace, as before.

The alias branch is the same three lines `binding` uses a few hundred lines below in the same file, for the same reason.

The failure is also split in two, since the two causes want different fixes:

```
defmethod: g is a Long, not a multimethod          ;; wrong var
defmethod: nope/area is not defined (no multimethod to extend)   ;; usually a missing require
```

## How it was found

Loading a real cross-namespace codebase under cljrs — [plato](https://github.com/BuddhiLW/plato), a presentation framework whose content renderer is a multimethod extended from four namespaces. With this fix its scene graph, timeline and SVG renderer all run on clojurust; before it, the whole tree failed to load.

## Tests

`crates/cljrs-runtime/tests/defmethod_cross_ns.rs`, 6 tests: through an alias, by full namespace, several namespaces extending one multimethod plus its `:default`, an unqualified name still resolving in the current namespace, and each of the two error messages.

`cargo test -p cljrs-runtime -p cljrs-stdlib` is clean.

### DIFF
--- a/crates/cljrs-runtime/src/interp/special.rs
+++ b/crates/cljrs-runtime/src/interp/special.rs
@@ -7,8 +7,7 @@ use crate::builtins::form::{
     expand_pairs, expand_reader_conds, expand_reader_conds_cow, form_to_value, resolve_auto_forms,
     select_reader_cond,
 };
-use crate::env::env::GlobalEnv;
-use crate::env::env::{Env, RequireRefer, RequireSpec};
+use crate::env::env::{Env, GlobalEnv, RequireRefer, RequireSpec};
 use crate::env::error::{EvalError, EvalResult};
 use crate::env::loader::load_ns;
 use crate::interp::destructure::bind_pattern;
@@ -2334,6 +2333,20 @@ fn eval_defmethod(args: &[Form], env: &mut Env) -> EvalResult {
             .unwrap_or_else(|| Arc::from(ns_part)),
         None => env.current_ns.clone(),
     };
+    // A pin can also live in the NAMESPACE half. `(require '[mylib@abc1234 :as
+    // v1])` registers the namespace under its literal versioned name, so the
+    // alias resolves straight to it and `parsed.version` is None; the guard
+    // above never sees it. Gated on the boundary like the privacy check below,
+    // so a versioned namespace's own source can still extend its own
+    // multimethods while it loads.
+    if owner_ns.as_ref() != env.current_ns.as_ref()
+        && cljrs_value::symbol::split_version(&owner_ns).1.is_some()
+    {
+        return Err(EvalError::Runtime(format!(
+            "defmethod: {multi_name} resolves to the versioned namespace {owner_ns}; \
+             extend the multimethod at HEAD"
+        )));
+    }
     // Reaching into another namespace respects privacy the way `eval_symbol`
     // does: refusing to extend a private multimethod from outside is what
     // `^:private` on a `defmulti` is for.
--- a/crates/cljrs-runtime/src/interp/special.rs
+++ b/crates/cljrs-runtime/src/interp/special.rs
@@ -7,6 +7,7 @@ use crate::builtins::form::{
     expand_pairs, expand_reader_conds, expand_reader_conds_cow, form_to_value, resolve_auto_forms,
     select_reader_cond,
 };
+use crate::env::env::GlobalEnv;
 use crate::env::env::{Env, RequireRefer, RequireSpec};
 use crate::env::error::{EvalError, EvalResult};
 use crate::env::loader::load_ns;
@@ -2317,6 +2318,15 @@ fn eval_defmethod(args: &[Form], env: &mut Env) -> EvalResult {
     // only in the current ns — otherwise `(defmethod other/render :x ...)`
     // reports that a perfectly good multimethod "is not a multimethod".
     let parsed = cljrs_value::Symbol::parse(multi_name);
+    // A pinned name (`mylib/render@abc1234`) refers to an immutable past
+    // version, which is not a thing a method can be installed into. `parse`
+    // splits the suffix off into `version`, so without this guard the pin is
+    // dropped and HEAD is extended instead.
+    if parsed.version.is_some() {
+        return Err(EvalError::Runtime(format!(
+            "defmethod: {multi_name} is versioned; extend the multimethod at HEAD"
+        )));
+    }
     let owner_ns: Arc<str> = match parsed.namespace.as_deref() {
         Some(ns_part) => env
             .globals
@@ -2324,12 +2334,24 @@ fn eval_defmethod(args: &[Form], env: &mut Env) -> EvalResult {
             .unwrap_or_else(|| Arc::from(ns_part)),
         None => env.current_ns.clone(),
     };
+    // Reaching into another namespace respects privacy the way `eval_symbol`
+    // does: refusing to extend a private multimethod from outside is what
+    // `^:private` on a `defmulti` is for.
+    if owner_ns.as_ref() != env.current_ns.as_ref()
+        && let Some(var) = env.globals.lookup_var(&owner_ns, &parsed.name)
+        && GlobalEnv::var_is_private(var.get())
+    {
+        return Err(EvalError::Runtime(format!(
+            "defmethod: var {owner_ns}/{} is not public",
+            parsed.name
+        )));
+    }
 
     let mf_ptr = match env.globals.lookup_in_ns(&owner_ns, &parsed.name) {
         Some(Value::MultiFn(mf)) => mf,
         Some(other) => {
             return Err(EvalError::Runtime(format!(
-                "defmethod: {multi_name} is a {}, not a multimethod",
+                "defmethod: {multi_name} is bound to a {}, not a multimethod",
                 other.type_name()
             )));
         }
--- a/crates/cljrs-runtime/src/interp/special.rs
+++ b/crates/cljrs-runtime/src/interp/special.rs
@@ -2310,12 +2310,32 @@ fn eval_defmethod(args: &[Form], env: &mut Env) -> EvalResult {
     let (multi_name, _) = require_sym_meta(args, 0, "defmethod", env)?;
     let multi_name = multi_name.as_str();
 
-    let mf_ptr = match env.globals.lookup_in_ns(&env.current_ns, multi_name) {
+    // The multimethod may live in another namespace, which is the normal case
+    // for an open dispatch: one namespace owns the `defmulti`, others extend
+    // it. Resolve `alias/name` and `fully.qualified.ns/name` the way ordinary
+    // qualified symbols resolve (`eval_symbol`, `binding`), instead of looking
+    // only in the current ns — otherwise `(defmethod other/render :x ...)`
+    // reports that a perfectly good multimethod "is not a multimethod".
+    let parsed = cljrs_value::Symbol::parse(multi_name);
+    let owner_ns: Arc<str> = match parsed.namespace.as_deref() {
+        Some(ns_part) => env
+            .globals
+            .resolve_alias(&env.current_ns, ns_part)
+            .unwrap_or_else(|| Arc::from(ns_part)),
+        None => env.current_ns.clone(),
+    };
+
+    let mf_ptr = match env.globals.lookup_in_ns(&owner_ns, &parsed.name) {
         Some(Value::MultiFn(mf)) => mf,
-        _ => {
+        Some(other) => {
             return Err(EvalError::Runtime(format!(
-                "defmethod: {} is not a multimethod",
-                multi_name
+                "defmethod: {multi_name} is a {}, not a multimethod",
+                other.type_name()
+            )));
+        }
+        None => {
+            return Err(EvalError::Runtime(format!(
+                "defmethod: {multi_name} is not defined (no multimethod to extend)"
             )));
         }
     };
--- a/crates/cljrs-runtime/tests/defmethod_cross_ns.rs
+++ b/crates/cljrs-runtime/tests/defmethod_cross_ns.rs
@@ -150,3 +150,80 @@ fn extending_something_undefined_says_that_instead() {
     let err = eval_in(&mut env, "(defmethod nope/area :a [_] 1)").expect_err("no such multimethod");
     assert!(err.contains("not defined"), "unhelpful error: {err}");
 }
+
+#[test]
+fn a_multimethod_reached_through_refer_needs_no_qualification() {
+    // The one path where the unqualified branch does real work rather than
+    // just preserving same-ns behaviour: `:refer` puts the name in the
+    // current namespace, so a bare `defmethod` has to find it there.
+    let extender = r#"
+(ns shapes.referred (:require [shapes.core :refer [area]]))
+(defmethod area :triangle [t] (:base t))
+"#;
+    let (_dir, _g, mut env) = env_with_sources(&[
+        ("shapes/core.cljrs", OWNER),
+        ("shapes/referred.cljrs", extender),
+    ]);
+    eval_in(&mut env, "(require 'shapes.referred)").expect("require the extending ns");
+    let v = eval_in(
+        &mut env,
+        "(do (require '[shapes.core :as c]) (c/area {:kind :triangle :base 9}))",
+    )
+    .expect("dispatch");
+    assert_eq!(v, Value::Long(9));
+}
+
+#[test]
+fn a_private_multimethod_cannot_be_extended_from_another_namespace() {
+    // `^:private` is refused here for the same reason `eval_symbol` refuses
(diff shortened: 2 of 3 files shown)


crates/cljrs-runtime/src/interp/special.rs
```diff
@@ -2310,12 +2310,32 @@ fn eval_defmethod(args: &[Form], env: &mut Env) -> EvalResult {
...
+    let parsed = cljrs_value::Symbol::parse(multi_name);
+    let owner_ns: Arc<str> = match parsed.namespace.as_deref() {
+        Some(ns_part) => env
+            .globals
+            .resolve_alias(&env.current_ns, ns_part)
+            .unwrap_or_else(|| Arc::from(ns_part)),
+        None => env.current_ns.clone(),
+    };
+
+    let mf_ptr = match env.globals.lookup_in_ns(&owner_ns, &parsed.name) {
```
csm, line 2363: Privacy is no longer enforced on the cross-namespace path. `eval_symbol` refuses a private var from another ns (eval.rs:282-290, `var: {ns}/{name} is not public`), but `defmethod` now extends one silently. Verified against this branch:

```clojure
;; shapes/core.cljrs
(ns shapes.core)
(defmulti ^:private secret :kind)

;; shapes/pv.cljrs
(ns shapes.pv (:require [shapes.core :as core]))
(defmethod core/secret :x [_] 1)     ;; => Ok(Nil), method installed
```

Before this PR that failed (for the wrong reason — "is not a multimethod"); Clojure throws `var: shapes.core/secret is not public`. The check that `eval_symbol` already uses drops in unchanged:

```rust
if owner_ns.as_ref() != env.current_ns.as_ref()
    && let Some(var) = env.globals.lookup_var(&owner_ns, &parsed.name)
    && GlobalEnv::var_is_private(var.get())
{
    return Err(EvalError::Runtime(format!(
        "defmethod: var {owner_ns}/{} is not public",
        parsed.name
    )));
}
```

---

crates/cljrs-runtime/src/interp/special.rs
```diff
@@ -2310,12 +2310,32 @@ fn eval_defmethod(args: &[Form], env: &mut Env) -> EvalResult {
...
     let multi_name = multi_name.as_str();
 
-    let mf_ptr = match env.globals.lookup_in_ns(&env.current_ns, multi_name) {
+    // The multimethod may live in another namespace, which is the normal case
+    // for an open dispatch: one namespace owns the `defmulti`, others extend
+    // it. Resolve `alias/name` and `fully.qualified.ns/name` the way ordinary
+    // qualified symbols resolve (`eval_symbol`, `binding`), instead of looking
+    // only in the current ns — otherwise `(defmethod other/render :x ...)`
+    // reports that a perfectly good multimethod "is not a multimethod".
+    let parsed = cljrs_value::Symbol::parse(multi_name);
```
csm, line 2319: `Symbol::parse` splits off a `@hash` version suffix into `parsed.version`, and this uses `parsed.name`, so the suffix is now silently dropped: `(defmethod mylib/render@abc1234 :x ...)` extends HEAD's `mylib/render` instead of erroring as it did before. `eval_symbol` routes versioned symbols to `resolve_versioned_symbol` rather than ignoring the pin (eval.rs:224-234).

Extending a pinned var isn't meaningful, so rejecting is the right answer — one guard before the lookup:

```rust
if parsed.version.is_some() {
    return Err(EvalError::Runtime(format!(
        "defmethod: {multi_name} is versioned; extend the multimethod at HEAD"
    )));
}
```

---

crates/cljrs-runtime/src/interp/special.rs
```diff
@@ -2310,12 +2310,32 @@ fn eval_defmethod(args: &[Form], env: &mut Env) -> EvalResult {
...
+            .unwrap_or_else(|| Arc::from(ns_part)),
+        None => env.current_ns.clone(),
+    };
+
+    let mf_ptr = match env.globals.lookup_in_ns(&owner_ns, &parsed.name) {
         Some(Value::MultiFn(mf)) => mf,
-        _ => {
+        Some(other) => {
+            return Err(EvalError::Runtime(format!(
+                "defmethod: {multi_name} is a {}, not a multimethod",
```
csm: Nit: `Value::type_name` returns lowercase names, so this reads `defmethod: g is a long, not a multimethod` (not `is a Long` as the PR description shows), and `(def h nil)` gives `defmethod: h is a nil, not a multimethod`. Something like `is bound to a {}` / `{} is a long, not a multimethod` avoids the "a nil" article. The tests only assert on the `not a multimethod` substring, so either wording passes.

---

crates/cljrs-runtime/tests/defmethod_cross_ns.rs
csm, line 1: `CLAUDE.md` requires the crate README to list every source file, and `crates/cljrs-runtime/README.md` does enumerate `tests/` (lines 105-130) — including the closely analogous `qualified_protocol_impl.rs` ("a qualified protocol name in an impl position … resolves through its own namespace"). This file needs an entry there in the same commit, e.g.

```
  defmethod_cross_ns.rs            — `defmethod` on a multimethod owned by
                                     another namespace, named through a
                                     `:require :as` alias or in full
```

Separately, a case worth adding since the fix reaches it: extending through `(:require [shapes.core :refer [area]])` with a bare `(defmethod area ...)`. It works on this branch (`lookup_in_ns` checks refers), but nothing pins it, and it is the one path where the `None => env.current_ns` branch has to do real work rather than just preserving same-ns behaviour.

---

csm (commented):
Reviewed the diff and exercised the branch locally.

**The fix is correct and the diagnosis is right.** `lookup_in_ns` keys interns and refers by bare name, so the old `lookup_in_ns(&current_ns, "core/area")` could never match a slash-bearing string — a qualified `defmethod` target was unreachable by construction, not just in edge cases. Resolving alias → full ns → current ns matches `eval_symbol` (eval.rs:262-295) and `binding` (special.rs:2395-2410). Mutating `mf_ptr.get().methods` behind the `GcPtr` reaches the owner's var, which is what makes the open dispatch actually open. `defmethod` is interpreter-only in every execution mode (rejected by `anf.rs:736`, forced to the interpreter by `aot.rs:1364`), so this one site covers the tree-walker, IR and AOT paths — no parallel fix needed. Splitting the two error causes is a genuine improvement; "not defined" pointing at a missing require is the common case.

Verified locally on `1669b9b`: the 6 new tests pass, the full `cargo test -p cljrs-runtime` suite is green (no regressions), `cargo fmt --check` and `cargo clippy -p cljrs-runtime --all-targets` are clean. I also checked that no pre-existing behaviour depended on the old lookup — nothing interns a slash-bearing name — and that `remove-method` / `prefer-method` have no equivalent gap, since they take the multifn as an evaluated argument and so already resolve through `eval_symbol`.

Four things, one of which I'd want fixed before merge:

1. **Privacy is no longer enforced** on the new cross-ns path — `(defmethod other.ns/private-mm ...)` now succeeds silently where `eval_symbol` would refuse the same symbol. Inline, with the check to reuse.
2. **Versioned names are silently unpinned** — `Symbol::parse` strips `@hash` into `parsed.version` and this reads `parsed.name`. Narrow, but it turns a previous error into a silent wrong-target. Inline.
3. **README not updated** — `CLAUDE.md` requires it and `crates/cljrs-runtime/README.md` enumerates `tests/`. Inline, with a suggested entry.
4. Wording nit on the type name in the error message, plus a suggested test for the `:refer` path.

The three-line alias resolution is now in its sixth copy (eval.rs:267, special.rs:875/2323/2401/2700, macros.rs:207, syntax_quote.rs:236). Not this PR's job to fix, but an `Env::resolve_ns_part(&self, ns_part) -> Arc<str>` would be worth a follow-up.

---

BuddhiLW:
Both holes are real and both were hidden rather than handled by the narrow lookup. Fixed in 9bce6a8.

**Privacy.** Your check, applied unchanged. Worth stating why it was safe to omit before and isn't now: with `lookup_in_ns(current_ns, "core/area")` the cross-ns path was unreachable by construction, so the privacy question never arose at this site — widening the lookup is exactly what makes it arise. Two tests: refused across the boundary, still allowed for the owning namespace, since privacy is about the boundary rather than the var.

**Version pins.** Refused rather than silently unpinned. Installing a method into an immutable past version isn't meaningful, so erroring is the honest answer, and it restores the pre-PR behaviour of not quietly hitting a different target.

**Error wording.** Now "is bound to a {}", which reads correctly for `a long` and `a nil`.

**README.** Entry added in the same commit. I also added the `:refer` test you suggested — you were right that it's the one path where the `None => current_ns` branch does real work rather than preserving same-ns behaviour, and nothing pinned it.

On the sixth copy of the alias resolution: agreed, and I'd like to take that on separately rather than smuggle it into this PR. It's the specific shape of a broader thing — every special form hand-written in Rust re-derives namespace resolution, and now privacy and version handling too, which is why this PR needed two follow-up guards that `eval_symbol` already had. An `Env::resolve_ns_part` collapses the first; the rest of the answer is fewer special forms in Rust. I'll open that as its own PR against the sites you listed.

crates/cljrs-runtime/src/interp/special.rs
```diff
@@ -2310,12 +2311,53 @@ fn eval_defmethod(args: &[Form], env: &mut Env) -> EvalResult {
...
+        )));
+    }
+    let owner_ns: Arc<str> = match parsed.namespace.as_deref() {
+        Some(ns_part) => env
+            .globals
+            .resolve_alias(&env.current_ns, ns_part)
+            .unwrap_or_else(|| Arc::from(ns_part)),
+        None => env.current_ns.clone(),
+    };
+    // Reaching into another namespace respects privacy the way `eval_symbol`
```
csm, line 2350: The version guard closes the `name@hash` spelling, but a pin can also live in the **namespace** half, and there the alias hides it from `parsed.version` completely.

`(require '[mylib@abc1234 :as v1])` is a supported form (`RequireSpec.version`, exercised at `tests/versioned_resolution.rs:193`). `load_versioned_ns` registers the namespace under the literal name `"mylib@abc1234"` (`versioned.rs:197`) and `apply_alias_refer` points the alias straight at it (`loader.rs:239`). So by the time `owner_ns` is computed, the pin is inside `owner_ns` and `parsed.version` is `None` — the guard above never fires.

Reproduced on `9bce6a8` by standing up exactly the state a versioned require leaves behind (a namespace registered as `mylib@abc1234`, plus `add_alias("user", "v1", "mylib@abc1234")`):

```
A) (defmethod v1/render :x [_] ...)              => Ok  ← installs :x into the pinned multimethod
B) (defmethod mylib@abc1234/render :y [_] ...)   => Err ... is versioned
C) (defmethod v1/render@abc1234 :z [_] 1)        => Err ... is versioned
```

The returned MultiFn in (A) carries the new `:x` alongside the `:default` whose `defining_ns` is `mylib@abc1234` — same object, mutated. That is the thing this commit set out to refuse, reached by the ordinary spelling.

The check wants to go after alias resolution, and gated on the boundary the same way the privacy check is — a versioned namespace's own source must still be able to `defmethod` into itself while it loads (`Env::new_versioned` sets `current_ns` to `"base@commit"`):

```rust
// The pin can also live in the namespace half (`(require '[mylib@abc1234
// :as v1])` aliases `v1` to the namespace `mylib@abc1234`), where the
// alias hides it from `parsed.version` entirely.
if owner_ns.as_ref() != env.current_ns.as_ref()
    && cljrs_value::symbol::split_version(&owner_ns).1.is_some()
{
    return Err(EvalError::Runtime(format!(
        "defmethod: {multi_name} resolves to the versioned namespace {owner_ns}; \
         extend the multimethod at HEAD"
    )));
}
```

`split_version` rather than `versioned::base_ns_name` keeps this out of the `#[cfg(not(target_arch = "wasm32"))]` module, so the WASM build is unaffected. I applied this locally: (A) becomes `defmethod: v1/render resolves to the versioned namespace mylib@abc1234; extend the multimethod at HEAD`, and the full `cargo test -p cljrs-runtime` suite stays green (52/52 binaries), `versioned_resolution.rs` included.

Unrelated nit on the same hunk: `use crate::env::env::GlobalEnv;` can fold into the `use crate::env::env::{Env, RequireRefer, RequireSpec};` line right below it.

---

csm (commented):
All four points from the last round are addressed, and addressed correctly.

**Privacy** — the check is in the right place, gated on `owner_ns != current_ns`, and uses `lookup_var` (interns only), which is what you want here: a var reachable in `owner_ns` only as a *refer* is necessarily public, since `refer_named`/`refer_all` both skip private vars (env.rs:440, env.rs:672). So there is no way in through `(ns a (:require [b :refer [mm]]))` + `(defmethod a/mm ...)`. The companion test for the owner extending its own private multimethod is the right pair to have.

**`:refer` test** — good, and it does pin the one path where the unqualified branch has real work to do.

**README** — matches the surrounding style and the file it sits next to.

**Message wording** — "is bound to a long" reads correctly now.

Verified on `9bce6a8`: all 10 tests in the file pass, full `cargo test -p cljrs-runtime` green with no regressions, `cargo fmt --check` and `cargo clippy -p cljrs-runtime --all-targets` clean. CI is green.

One remaining gap, inline: the version guard covers `name@hash` but not a pin carried in the *namespace* half. `(require '[mylib@abc1234 :as v1])` aliases `v1` directly to the namespace `mylib@abc1234`, so `(defmethod v1/render :x ...)` has `parsed.version == None` and installs a method into the pinned multimethod. Reproduced, with a four-line fix that goes after alias resolution and keeps the whole suite green.

---

BuddhiLW:
Fixed in `b8a217e` (rebased onto the `main` merge already on the branch).

You reproduced it exactly, and the framing is the useful part: the pin lives in *either half of the name*, and only one half reaches `Symbol::parse`. Once the alias resolves, the pin is inside `owner_ns` and `parsed.version` is `None`, so the earlier guard is looking at the wrong string. Your check applied as given, after alias resolution and gated on the boundary:

```rust
if owner_ns.as_ref() != env.current_ns.as_ref()
    && cljrs_value::symbol::split_version(&owner_ns).1.is_some()
{
    return Err(EvalError::Runtime(format!(
        "defmethod: {multi_name} resolves to the versioned namespace {owner_ns}; \
         extend the multimethod at HEAD"
    )));
}
```

`split_version` over `versioned::base_ns_name` for the reason you gave — it keeps this out of the `#[cfg(not(target_arch = "wasm32"))]` module.

Two tests, standing up the same state a versioned require leaves behind (namespace registered as `mylib@abc1234`, `add_alias("user", "v1", "mylib@abc1234")`):

- **the refusal**, and — the part worth asserting separately — that the pinned method table is *untouched* afterwards. `(v1/render {:kind :x})` still answers `:from-the-pin`, so the guard fires before the mutation rather than after it.
- **the owner extending itself**, which is what the boundary gate buys: `Env::new_versioned` sets `current_ns` to the versioned name, so an ungated check would break the loading path the pin exists to serve.

Mutation-checked: neutering the condition fails `a_pin_carried_in_the_namespace_half_is_refused_too` and nothing else.

Import nit folded in the same commit.

### One thing I found and deliberately did not change

A qualified *self*-reference inside a versioned namespace now hits the older `parsed.version` guard:

```clojure
;; while mylib@abc1234 is loading, current_ns == "mylib@abc1234"
(defmethod mylib@abc1234/render :y [_] ...)
;; => defmethod: mylib@abc1234 is versioned; extend the multimethod at HEAD
```

The unqualified form works and is what the loader actually emits, so nothing is broken today. But `resolve_versioned_value` explicitly supports "a qualified self-reference inside an already-versioned namespace" (`env/versioned.rs:59-61`), so `defmethod` and symbol resolution disagree about that spelling. That guard predates this round and the fix is not obviously one line — it wants the same `owner_ns != current_ns` gate the new check has, which changes behaviour the previous round's test pins. Flagging rather than folding it in; say the word and I will take it either way.

`cargo test -p cljrs-runtime` green (54 binaries), `cargo fmt --all --check` and `cargo clippy -p cljrs-runtime --all-targets` clean.

PR #379, opened 2026-09-08
Load a native extension from a working tree (:local/root + :rust/load :dylib)

BuddhiLW:
`:rust/load :dylib` could only name a git dep at a pinned commit, so a Rust extension had to be committed and pushed before the runtime could load it even once. The `:rust/*` keys were already parsed on a `:local/root` dep and then dropped on the floor — `Dependency::Local` had nowhere to put them, and `find_dylib_dep` matched only `Dependency::Git`.

A local root now carries the same keys and builds the working tree as it stands:

```clojure
{:deps {my.lib {:local/root "../my-lib"
                :rust/load  :dylib
                :rust/init  "my_lib::cljrs_init"
                :rust/crate "src/crates/thing"}}}
```

### Shape

The two source forms differ only in where the crate comes from, so they meet at a `NativeDep` and share one build path: materialize the source, identify the version, plan the paths, build, load.

Identifying the version is what makes a working tree safe to cache. A commit names a pinned build; a digest of the crate's source files names a local one. Both are immutable, so the artifact cache is trustworthy in both cases and an edited tree gets its own wrapper directory.

That last part is load-bearing rather than an optimization. A rebuilt library reusing the old path is **not** picked up: the process has already `dlopen`ed that path and gets the existing mapping back. The e2e test caught exactly that — it asserted an edit was visible, and the first implementation returned the stale value.

A local dep is declined for versioned (`ns/f@sha`) resolution: it has no commit, so it cannot answer for one, and the resolver falls back exactly as before.

### Second commit: a naming bug this exposed

The generated wrapper declared its dependency under the first segment of the `:rust/init` path — a Rust *identifier* — with a comment asserting Cargo would read the real package name from the crate's own Cargo.toml. It does not; the key of a path dependency is the package name being looked up. Any package whose name contains `-` failed:

```
error: no matching package named `cljrs_raster` found
help: packages with similar names: cljrs-raster
```

The dependency is now declared under the ident the init path uses and renamed to the real `[package] name`. This affects git deps identically — the existing fixture is called `pinlib`, which has no hyphen, so nothing covered the claim.

### Tests

Five parser cases for the new keys; unit tests for the manifest read (workspace-inherited version, a `name` in another table); two gated e2e cases for build-from-tree, rebuild-after-edit, and the versioned fallback. `CLJRS_DYLIB_E2E=1 cargo test -p cljrs --test pinned_dylib_e2e` passes 4/4; `cargo test -p cljrs -p cljrs-project` is green.

Docs: a "Native dependencies" section in `cli/deps.md`, including the two limits that follow from a local dep having no commit — no versioned symbols, and no reproducibility.

### DIFF
--- a/crates/cljrs/src/native/pinned.rs
+++ b/crates/cljrs/src/native/pinned.rs
@@ -1,24 +1,35 @@
-//! Pinned native packages: build a dependency's Rust crate at a pinned git
-//! commit as a cdylib and load it (`:rust/load :dylib` in `cljrs.edn`).
+//! Native packages: build a dependency's Rust crate as a cdylib and load it
+//! (`:rust/load :dylib` in `cljrs.edn`).
+//!
+//! Two dependency forms reach this module, and they differ only in where the
+//! crate's source comes from:
+//!
+//! - `:git/url` + `:git/sha` — an immutable commit. Serves both `require` and
+//!   versioned (`mylib/f@<sha>`) resolution; its artifact is cached forever.
+//! - `:local/root` — a working tree, built as it currently stands. Serves
+//!   `require` only (there is no commit to resolve `@<sha>` against), and is
+//!   rebuilt on every load because the tree changes underfoot.
 //!
 //! By default, a pinned symbol (`mylib/f@<sha>`) that resolves to a native
 //! (Rust-backed) function falls back to the **current binary's**
 //! implementation, with provenance verification (see
-//! `cljrs_runtime::env::versioned`).  This crate provides the opt-in alternative:
-//! true pinned native code.
+//! `cljrs_runtime::env::versioned`).  This module provides the opt-in
+//! alternative: true out-of-binary native code.
 //!
 //! ## Flow
 //!
-//! 1. `install(globals)` registers a [`PinnedNativeLoader`] hook on the
-//!    environment.  The versioned resolver calls it whenever a pinned
-//!    lookup is about to fall back to a native function.
+//! 1. `install(globals)` registers the loader hooks on the environment.  The
+//!    versioned resolver calls one whenever a pinned lookup is about to fall
+//!    back to a native function; `require` calls the other for a namespace no
+//!    source path provides.
 //! 2. The hook checks `cljrs.edn` for a dep covering the namespace with
 //!    `:rust/load :dylib` and a `:rust/init` function.
-//! 3. The dep's repository is fetched (`cljrs_project::vcs::fetch_remote`), checked
-//!    out at the pinned commit, and wrapped in a generated cdylib crate
-//!    that pins the exact same `cljrs-interop` as the host.
-//! 4. `cargo build --release` (cached per `(crate, commit, rustc, cljrs
-//!    version)`), then `dlopen`.
+//! 3. The dep's source is materialized — fetched and checked out at the
+//!    pinned commit (`cljrs_project::vcs`), or taken as-is from
+//!    `:local/root` — and wrapped in a generated cdylib crate that pins the
+//!    exact same `cljrs-interop` as the host.
+//! 4. `cargo build` in the host's profile (cached per `(crate, source, rustc,
+//!    cljrs version)`), then `dlopen`.
 //! 5. **ABI handshake**: the wrapper exports `cljrs_dylib_abi()` returning
 //!    a fingerprint string (cljrs version + `rustc -V`, baked at the
 //!    wrapper's build time).  The host refuses to proceed unless it equals
@@ -44,7 +55,7 @@
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
 
-use cljrs_project::config::{Dependency, GitDep};
+use cljrs_project::config::Dependency;
 use cljrs_runtime::env::error::{EvalError, EvalResult};
 use cljrs_runtime::tiered::GlobalEnv;
 
@@ -92,16 +103,22 @@ fn load_pinned(globals: &Arc<GlobalEnv>, base_ns: &str, commit: &str) -> EvalRes
     let Some(config) = config else {
         return Ok(false);
     };
-    let Some(git) = find_dylib_dep(&config, base_ns) else {
+    let Some(dep) = find_dylib_dep(&config, base_ns) else {
         return Ok(false);
     };
+    let dep = dep.map_err(|e| EvalError::Runtime(format!("pinned native {base_ns}: {e}")))?;
+    // A working tree has no commit, so it cannot answer for `ns/f@sha`.
+    // Decline and let the versioned resolver fall back as it would have.
+    if !dep.is_pinned() {
+        return Ok(false);
+    }
 
     let versioned_ns = format!("{base_ns}@{commit}");
     if globals.is_loaded(&versioned_ns) {
         return Ok(true);
     }
 
-    let lib_path = build_pinned_wrapper(&git, commit)
+    let lib_path = build_wrapper(&dep, Some(commit))
         .map_err(|e| EvalError::Runtime(format!("pinned native {versioned_ns}: {e}")))?;
 
     load_library(globals, &lib_path, Some(commit))
@@ -113,22 +130,26 @@ fn load_pinned(globals: &Arc<GlobalEnv>, base_ns: &str, commit: &str) -> EvalRes
 }
 
 /// The `require`-path loader hook: returns `Ok(true)` when a `:rust/load
-/// :dylib` dep covering `ns` was built at its pinned `:git/sha` and its
-/// exports were registered into the **unversioned** namespace, making a plain
-/// `(require '[ns :as …])` of a pure-native package succeed.
+/// :dylib` dep covering `ns` was built and its exports were registered into
+/// the **unversioned** namespace, making a plain `(require '[ns :as …])` of a
+/// pure-native package succeed.
 ///
 /// Unlike [`load_pinned`] (which serves versioned-symbol resolution and lands
 /// the package in the immutable `"<ns>@<commit>"` namespace), this registers
 /// into the live namespace so unversioned references resolve normally.  The
 /// caller (`cljrs-runtime`'s unversioned loader) marks `ns` loaded on success.
+///
+/// Both dependency forms reach this path: a git dep builds at its pinned
+/// `:git/sha`, a `:local/root` dep builds the working tree as it stands.
 fn load_require(globals: &Arc<GlobalEnv>, ns: &str) -> EvalResult<bool> {
     let config = globals.deps_config.read().unwrap().clone();
     let Some(config) = config else {
         return Ok(false);
     };
-    let Some(git) = find_dylib_dep(&config, ns) else {
+    let Some(dep) = find_dylib_dep(&config, ns) else {
         return Ok(false);
     };
+    let dep = dep.map_err(|e| EvalError::Runtime(format!("native dep {ns}: {e}")))?;
 
     // Already brought in (e.g. an earlier require of a sibling namespace
     // provided by the same dylib loaded the whole package).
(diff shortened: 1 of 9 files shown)


crates/cljrs/src/native/pinned.rs
```diff
@@ -113,139 +130,393 @@ fn load_pinned(globals: &Arc<GlobalEnv>, base_ns: &str, commit: &str) -> EvalRes
...
+            if name == "target" || name == ".git" {
+                continue;
+            }
+            let path = entry.path();
+            if path.is_dir() {
+                stack.push(path);
+                continue;
+            }
+            let rel = path.strip_prefix(dir).unwrap_or(&path);
+            let bytes = std::fs::read(&path).map_err(|e| format!("reading {}: {e}", path.display()))?;
```
csm: **Blocking: this is the CI failure.** `cargo fmt --check` rejects this line (over the width limit), and it's the only thing red in the "Build, test, lint" job — clippy and the unit tests are clean.

```diff
-            let bytes = std::fs::read(&path).map_err(|e| format!("reading {}: {e}", path.display()))?;
+            let bytes =
+                std::fs::read(&path).map_err(|e| format!("reading {}: {e}", path.display()))?;
```

---

crates/cljrs/src/native/pinned.rs
```diff
@@ -113,139 +130,393 @@ fn load_pinned(globals: &Arc<GlobalEnv>, base_ns: &str, commit: &str) -> EvalRes
...
+    match source {
+        NativeSource::Pinned { url, sha } => {
+            let commit = commit_override.unwrap_or(sha.as_ref());
+            Ok(SourceVersion {
+                slug: format!("@{commit}"),
+                key: format!("{url}|{commit}"),
+            })
+        }
+        NativeSource::WorkingTree(root) => {
+            let digest = digest_source_tree(crate_dir)?;
```
csm: **Blocking: an edit outside `:rust/crate` is silently not rebuilt.**

The digest covers `crate_dir`, not the local root. With `:rust/crate "src/crates/thing"` — the layout in both the PR description and `docs/book/src/cli/deps.md:182` — a multi-crate root normally has path deps on siblings. Edit a sibling crate and the digest is unchanged, so `build_wrapper` takes the `plan.artifact.exists()` early return and `dlopen`s the stale library. No rebuild happens at all; cargo never gets a chance to notice.

That's the exact failure the design section says the version-identity scheme exists to prevent ("A rebuilt library reusing the old path is **not** picked up"). The e2e test doesn't catch it because `write_pinlib_tree` puts a single crate at the root, where `crate_dir == source_root`.

Digesting `source_root` instead trades spurious rebuilds (any `.cljc` edit under the root re-digests) for correctness, which seems like the right side of the trade for a key whose whole job is "the tree as it stands". If you'd rather keep the narrow digest, the limit needs stating in the docs right next to the `:rust/crate` example, since that's where a reader will hit it.

---

crates/cljrs/src/native/pinned.rs
```diff
@@ -113,139 +130,393 @@ fn load_pinned(globals: &Arc<GlobalEnv>, base_ns: &str, commit: &str) -> EvalRes
...
+            if path.is_dir() {
+                stack.push(path);
+                continue;
+            }
+            let rel = path.strip_prefix(dir).unwrap_or(&path);
+            let bytes = std::fs::read(&path).map_err(|e| format!("reading {}: {e}", path.display()))?;
+            acc.push_str(&format!(
+                "{}|{}\n",
+                rel.display(),
+                stable_hash(&String::from_utf8_lossy(&bytes))
```
csm: `String::from_utf8_lossy` maps every invalid byte to U+FFFD, so two files differing only in invalid UTF-8 sequences digest identically and the stale artifact gets reused. Unlikely to bite in a Rust crate, but hashing the bytes directly costs nothing — a `stable_hash_bytes(&[u8])` alongside `stable_hash` would do it, since `Hasher::write` takes bytes anyway.

---

crates/cljrs/src/native/pinned.rs
```diff
@@ -113,139 +130,393 @@ fn load_pinned(globals: &Arc<GlobalEnv>, base_ns: &str, commit: &str) -> EvalRes
...
-    write_wrapper_crate(&wrapper_dir, &crate_dir, crate_name, &pkg_ident, init_fn)?;
-
-    // 4. Build, matching the host's profile (see `abi_fingerprint`).
-    let offline = find_workspace_root().is_some();
+/// Run `cargo build` on the generated wrapper, matching the host's profile
+/// (see [`abi_fingerprint`]).
+fn cargo_build(plan: &WrapperPlan) -> Result<(), String> {
     let mut cmd = std::process::Command::new("cargo");
-    cmd.arg("build").current_dir(&wrapper_dir);
+    cmd.arg("build").current_dir(&plan.wrapper_dir);
```
csm: **Every save costs a from-scratch build and leaks a target directory.**

`wrapper_dir` is `cache/dylibs/<pkg>@local-<digest>/fp-<hash>` and this runs with no `CARGO_TARGET_DIR`, so each new digest gets an empty `target/` and recompiles `cljrs-interop` and its whole dependency tree. There's no eviction anywhere under `dylib_cache_root()`, so an afternoon of edits leaves one full target dir per save on disk.

Both are avoidable without giving up the distinct-artifact-path property `dlopen` needs: build in a `CARGO_TARGET_DIR` shared per `(dep, fingerprint)` and copy the resulting cdylib out to the digest-unique path.

Separately, `--offline` on line 483 is keyed to "a cljrs checkout was found", which is about pinning `cljrs-interop` by path — it doesn't need to constrain the dep's own resolution. Pre-existing, but a locally-developed extension pulling something from crates.io that isn't already in the cargo cache will now fail to build for anyone running `cljrs` from a source checkout.

---

crates/cljrs/src/native/pinned.rs
```diff
@@ -1,24 +1,35 @@
...
+//! (`:rust/load :dylib` in `cljrs.edn`).
+//!
+//! Two dependency forms reach this module, and they differ only in where the
+//! crate's source comes from:
+//!
+//! - `:git/url` + `:git/sha` — an immutable commit. Serves both `require` and
+//!   versioned (`mylib/f@<sha>`) resolution; its artifact is cached forever.
+//! - `:local/root` — a working tree, built as it currently stands. Serves
+//!   `require` only (there is no commit to resolve `@<sha>` against), and is
+//!   rebuilt on every load because the tree changes underfoot.
```
csm: A local dep isn't rebuilt on every load — it's digest-cached, same as everything else the module does (`resolve_version` → `plan_wrapper` → the `artifact.exists()` early return). Worth correcting, since this is the first thing a reader meets and it describes the opposite policy from the one implemented.

---

crates/cljrs/src/native/pinned.rs
```diff
@@ -113,139 +130,393 @@ fn load_pinned(globals: &Arc<GlobalEnv>, base_ns: &str, commit: &str) -> EvalRes
...
+    let manifest = std::fs::read_to_string(crate_dir.join("Cargo.toml"))
+        .map_err(|e| format!("reading {}: {e}", crate_dir.join("Cargo.toml").display()))?;
+    let package_name = package_name_of(&manifest).ok_or_else(|| {
+        format!(
+            "no [package] name in {}",
+            crate_dir.join("Cargo.toml").display()
+        )
+    })?;
+    let dep_line = format!(
+        r#"{pkg_ident} = {{ path = "{}", package = "{package_name}" }}"#,
```
csm: The path goes into a TOML basic string unescaped (same for `interop_dep` below). A Windows path's backslashes are invalid escapes and a `"` in the path breaks the manifest outright. Pre-existing for git checkouts, but `:local/root` is the first time a user-chosen path lands here, and `wrapper_artifact_path` has a `target_os = "windows"` arm so Windows is nominally in scope. Escaping `\` and `"`, or emitting a literal string, fixes it.

---

crates/cljrs/src/native/pinned.rs
```diff
@@ -113,139 +130,393 @@ fn load_pinned(globals: &Arc<GlobalEnv>, base_ns: &str, commit: &str) -> EvalRes
...
+    }
+
+    /// Whether this dep names an immutable commit, and so can serve versioned
+    /// (`ns/f@sha`) resolution.
+    fn is_pinned(&self) -> bool {
+        self.source.commit().is_some()
+    }
+
+    /// The Cargo package name, as the init path's first segment spells it.
+    fn crate_name(&self) -> &str {
```
csm: The doc says "The Cargo package name", but the second commit's whole point is that the init path's first segment is *not* the package name — that comes from `package_name_of`. This returns the Rust identifier.

It's also only ever reached through `pkg_ident()`, whose `.replace('-', "_")` is dead: a Rust path segment can't contain `-` in the first place. Collapsing the two into `pkg_ident()` would remove both the contradiction and the no-op.

---

crates/cljrs/src/native/pinned.rs
```diff
@@ -92,16 +103,22 @@ fn load_pinned(globals: &Arc<GlobalEnv>, base_ns: &str, commit: &str) -> EvalRes
     let Some(config) = config else {
         return Ok(false);
     };
-    let Some(git) = find_dylib_dep(&config, base_ns) else {
+    let Some(dep) = find_dylib_dep(&config, base_ns) else {
         return Ok(false);
     };
+    let dep = dep.map_err(|e| EvalError::Runtime(format!("pinned native {base_ns}: {e}")))?;
```
csm: The `?` runs before the `is_pinned()` check below, so a `:local/root` dep with `:rust/load :dylib` and no `:rust/init` makes versioned resolution *error* rather than decline. The PR states a local dep "is declined for versioned (`ns/f@sha`) resolution … and the resolver falls back exactly as before" — a misconfigured one doesn't. Hoisting the `is_pinned()` check above this line restores the contract.

---

docs/book/src/cli/deps.md
```diff
@@ -152,3 +152,49 @@ vendor {:local/root "../vendor/utils"}
...
+
+Two limits apply to a local native dependency, both following from its having
+no commit:
+
+- It cannot serve a **versioned symbol** (`my.lib/f@<sha>`). Those resolve
+  against pinned dependencies only; a local one is skipped.
+- It is **not reproducible**. Two machines with different working trees get
+  different builds, silently. Only `:git/sha` makes a build repeatable.
+
+Builds are cached under `~/.cljrs/cache/dylib/` per crate, version, compiler
```
csm: `dylib_cache_root()` is `~/.cljrs/cache/dylibs/` (plural).

---

crates/cljrs/tests/pinned_dylib_e2e.rs
```diff
@@ -290,3 +290,179 @@ fn native_dep_loaded_by_plain_require() {
...
+        );
+        return;
+    }
+
+    let ws_root = workspace_root();
+    let tree = tempfile::tempdir().unwrap();
+    write_pinlib_tree(tree.path(), &ws_root, 7);
+
+    let cache_home = tempfile::tempdir().unwrap();
+    // SAFETY: single gated test invocation; no concurrent env readers.
```
csm: The `SAFETY` claim no longer holds: there are now four tests in this binary (two of them added here), each setting `HOME` to its own tempdir, and libtest runs them on parallel threads. `set_var` from multiple threads is UB, and the `HOME` values clobber each other.

The assertions still come out right — cache dirs are keyed by commit or digest, so cross-contamination is harmless in practice — but the comment asserts an invariant that's now false in all four places (lines 129, 231, 385, 428). Either serialize the tests or restate what the comment is actually claiming.

---

csm (changes requested):
Reviewed at `8724ab9`. Verified locally: `cargo clippy -p cljrs -p cljrs-project --all-targets` clean, `cargo test -p cljrs-project --lib` 22/22, `cargo test -p cljrs --lib native::pinned` 5/5. I did not run the gated e2e (`CLJRS_DYLIB_E2E=1`).

The `NativeSource` / `NativeDep` / `SourceVersion` / `WrapperPlan` split is a real improvement over the previous inlined `build_pinned_wrapper` — `materialize_source`, `resolve_version`, `plan_wrapper` and `cargo_build` each do one thing, and `plan_wrapper` is genuinely pure. The `package = "…"` rename in the second commit is correct, and the hyphen case it fixes was truly uncovered by the `pinlib` fixture. `package_name_of`'s tests hit the cases that matter (`[lib] name`, a dependency table's `name`, workspace-inherited `version`). Declining versioned resolution is the right call and is tested against a live host binding rather than just asserting absence.

Three things I'd want addressed before merge:

**1. CI is red on `cargo fmt --check`** (`pinned.rs:414`) — the only failure in the job; the WASM job is green. Inline comment has the diff.

**2. The digest misses sibling-crate edits under `:rust/crate`** (`pinned.rs:380`) — the cached artifact is returned and the stale library loaded, which is the failure the design section says the version scheme prevents. Details inline; the e2e fixture is single-crate, so it doesn't reach this.

**3. Stale crate READMEs**, which CLAUDE.md requires in the same commit:
- `crates/cljrs-project/README.md:128` still shows `Local { root: PathBuf }`; the variant has three new public fields.
- `crates/cljrs/README.md:304-336` ("Pinned native packages") describes only the git form and the two-commit fixture — no `:local/root`, no `digest_source_tree`, and the e2e tests it enumerates by name are now four, not two.

The rest of the inline comments are non-blocking: build/cache cost per edit, the lossy-UTF-8 digest, unescaped paths in the generated TOML, a doc/code contradiction at the top of the module, the `dylib`/`dylibs` doc typo, `crate_name()`'s now-wrong name, an error path that breaks the stated versioned-fallback contract, and the `SAFETY` comments in the e2e file that four parallel tests have invalidated.

---

crates/cljrs/src/native/pinned.rs
```diff
@@ -113,148 +133,496 @@ fn load_pinned(globals: &Arc<GlobalEnv>, base_ns: &str, commit: &str) -> EvalRes
...
+    fn from_env() -> NetworkPolicy {
+        NetworkPolicy::from_env_value(std::env::var("CLJRS_DYLIB_OFFLINE").ok().as_deref())
+    }
+}
 
-    // 4. Build, matching the host's profile (see `abi_fingerprint`).
-    let offline = find_workspace_root().is_some();
+/// Run `cargo build` on the generated wrapper, matching the host's profile
+/// (see [`abi_fingerprint`]).
+fn cargo_build(plan: &WrapperPlan) -> Result<(), String> {
```
csm, line 571: **The shared build directory reopens the stale-library hole through cargo's fingerprint.**

The version is content-based; the decision to actually recompile is now delegated to cargo, which is mtime-based. When they disagree, `build_output` is the *previous* build, `publish_artifact` copies it to the new version's path, and the stale library is `dlopen`ed — under a version that says it is fresh.

Cargo really does behave this way. Minimal repro:

```
$ echo 'pub const N: i64 = 1;' > src/lib.rs && cargo build -q
$ echo 'pub const N: i64 = 2;' > src/lib.rs && touch -d '2020-01-01' src/lib.rs
$ cargo build
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.01s
$ cargo build -v | grep Fresh
       Fresh mt v0.1.0 (...)
```

Content changed, digest changes, cargo does nothing. Triggers where mtime doesn't advance: `cp -p`, `rsync --times`, extracting a tarball that preserves times, restoring a file from a backup. Before this commit each version got its own build directory, so cargo compiled from scratch and couldn't be fooled — this is the cost of the sharing, and it lands squarely on the guarantee the digest exists to provide.

Cheapest fix I see: write the version that produced `build_output` into a marker file in `wrapper_dir` after a successful build. On a miss, if the marker already equals the version you're about to build, you have a cache-coherence bug; if after `cargo build` the marker's predecessor is still what's on disk and `build_output`'s mtime didn't move, the build didn't pick the change up — fail loudly rather than publishing it. Failing beats silently serving the old code, which is the outcome the whole design is arranged against.

---

crates/cljrs/src/native/pinned.rs
```diff
@@ -113,148 +133,496 @@ fn load_pinned(globals: &Arc<GlobalEnv>, base_ns: &str, commit: &str) -> EvalRes
...
+/// `dlopen` keys on the path, so a rebuilt library has to arrive somewhere the
+/// previous one never occupied; the shared build directory cannot provide
+/// that, and the copy is what buys it back.
+fn publish_artifact(plan: &WrapperPlan) -> Result<(), String> {
+    let dir = plan
+        .artifact
+        .parent()
+        .ok_or("wrapper artifact path has no parent directory")?;
+    std::fs::create_dir_all(dir).map_err(|e| format!("creating {}: {e}", dir.display()))?;
+    std::fs::copy(&plan.build_output, &plan.artifact).map_err(|e| {
```
csm, line 398: The copy isn't atomic, and the cache has no way to recover from a partial one. Ctrl-C mid-copy, a full disk, or the OOM killer leaves a truncated `.so` at `plan.artifact`; from then on `build_wrapper` takes the `artifact.exists()` early return at line 370 and hands it to `dlopen`, which fails. Nothing invalidates it, so every subsequent run fails the same way until someone works out to `rm -rf ~/.cljrs/cache/dylibs`.

Copy to a sibling temp path and `fs::rename` onto the final name — rename within a directory is atomic, so the artifact either isn't there or is complete. That also makes two processes publishing the same version concurrently safe, which the plain copy isn't.

---

crates/cljrs/src/native/pinned.rs
```diff
@@ -113,148 +133,496 @@ fn load_pinned(globals: &Arc<GlobalEnv>, base_ns: &str, commit: &str) -> EvalRes
...
+            .map_err(|e| format!("reading {}: {e}", current.display()))?;
+        entries.sort_by_key(|e| e.file_name());
+        for entry in entries {
+            let name = entry.file_name();
+            let name = name.to_string_lossy();
+            if name == "target" || name == ".git" {
+                continue;
+            }
+            let path = entry.path();
+            if path.is_dir() {
```
csm, line 463: Two things follow from the walk now covering the whole root rather than one crate directory — my ask, so the cost is mine too, but both are worth a look:

- `Path::is_dir()` resolves symlinks, so a symlink cycle anywhere under `:local/root` loops forever, growing `stack` until it OOMs. `entry.file_type()` doesn't follow, which both fixes the cycle and stops the walk from wandering out of the tree.
- Every byte of every file outside `target/` and `.git/` is read on every `require`. Fine for `../my-lib`; a `:local/root` pointing at a repo root with fixtures, `node_modules`, or another toolchain's build directory pays that on each process start. Skipping obvious non-inputs, or hashing `(len, mtime)` for files above some size, would bound it.

---

crates/cljrs/src/native/pinned.rs
```diff
@@ -113,148 +133,496 @@ fn load_pinned(globals: &Arc<GlobalEnv>, base_ns: &str, commit: &str) -> EvalRes
...
+    }
+}
+
+/// Derive every path the build needs. Pure: the source is already resolved.
+fn plan_wrapper(dep: &NativeDep, crate_dir: PathBuf, version: &SourceVersion) -> WrapperPlan {
+    let abi = abi_fingerprint();
+    let label = format!("{}{}", dep.pkg_ident(), version.slug);
+    let fp_hash = stable_hash(&format!("{abi}|{}", version.key));
+    let build_hash = stable_hash(&format!("{abi}|{}", crate_dir.display()));
     let wrapper_dir = dylib_cache_root()
```
csm, line 520: Nit, in passing: `write_wrapper_crate` runs on every artifact miss and rewrites `Cargo.toml`, `build.rs` and `src/lib.rs` with identical bytes but fresh mtimes, so cargo recompiles the wrapper crate and re-runs `build.rs` even when only the dep changed. Cheap next to the dep's own compile, and skipping the write when the content matches would remove it.

The accumulation is also much better than before — one cdylib per edit instead of one target tree — but still nothing prunes `~/.cljrs/cache/dylibs/<pkg>@local-*`. A long-lived working tree grows a directory per save forever. Not for this PR, but it will want an eviction pass eventually.

---

csm (commented):
Re-reviewed at `e0009a3` (new work is `6541b0d`; `e0009a3` and `162e31d` are merges). Every item from the last round is addressed, most of them with a test that would catch the regression:

| Last round | Now |
|---|---|
| `cargo fmt --check` red | Clean — verified locally |
| Digest scoped to `crate_dir`, sibling edits stale | Digests the whole `source_root`; unit test `a_working_tree_version_covers_the_whole_root` plus a real two-crate e2e (`…picks_up_a_sibling_crate_edit`) |
| Both crate READMEs stale | Both updated, and the `cljrs` one now documents the build/publish split and names all five e2e cases |
| Full rebuild + leaked target dir per edit | Shared `wrapper_dir` + explicit `CARGO_TARGET_DIR`, cdylib copied out to a version-unique path; `the_artifact_path_is_version_unique_but_the_build_dir_is_not` pins the invariant |
| Lossy-UTF-8 digest | `stable_hash_bytes`, with `a_digest_distinguishes_invalid_utf8` |
| Module doc claimed rebuild-every-load; `dylib`/`dylibs` typo | Both corrected |
| Unescaped paths in generated TOML | `toml_basic_string`, tested with a Windows path and an embedded quote |
| `crate_name()` misnamed, dead `replace('-', "_")` | Folded into `pkg_ident()` with the identifier/package-name distinction spelled out |
| Error before `is_pinned()` broke the fallback contract | `DylibDecl`/`validate` split; `a_local_dep_without_rust_init_declines_versioned_resolution` covers it and is deliberately ungated, since nothing should be built |
| False `SAFETY` comments, 4 tests racing on `HOME` | `ENV_LOCK` + `TestEnv`, with a `SAFETY` note that now states something true |
| `--offline` inferred from an unrelated flag | `NetworkPolicy` with `CLJRS_DYLIB_OFFLINE`, documented in `deps.md` and the README |

The `DylibDecl` → `validate` → `NativeDep` split is a better answer than what I suggested: making "can this serve `@<sha>`?" answerable before validation is what makes the decline-not-error behaviour structural rather than a matter of statement order.

Verified locally at this head: `cargo fmt --check` clean, `cargo clippy -p cljrs -p cljrs-project --all-targets` clean, `cargo test -p cljrs -p cljrs-project --lib` green (13/13 in `native::pinned`), and **`CLJRS_DYLIB_E2E=1 cargo test -p cljrs --test pinned_dylib_e2e` passes 6/6** (261s) — including the sibling-crate case, which fails against the previous head.

Nothing blocking. Four new comments inline, in rough order of how much I'd want them looked at:

1. **The shared build directory reopens the stale-library hole from the other side.** The version is content-based but the rebuild decision is now cargo's, which is mtime-based; when a content change doesn't advance mtime (`cp -p`, `rsync --times`, tarball extraction), cargo skips the build and the previous `build_output` gets published under the new version. Repro in the comment. Narrow trigger, but it's the exact failure the digest exists to prevent, and it wasn't reachable before this commit.
2. **`publish_artifact`'s copy isn't atomic**, so an interrupted first build poisons the cache permanently — `artifact.exists()` then serves a truncated `.so` to `dlopen` on every later run. `rename` fixes it.
3. **The widened digest walk follows symlinks** (cycle → unbounded loop) and reads every byte under the root on every `require`.
4. Nit: `write_wrapper_crate` rewrites identical bytes on each miss, forcing a wrapper recompile; and artifact directories still accumulate one per edit with no eviction.

---

csm (approved):
Approving. This also clears the `CHANGES_REQUESTED` I left at `8724ab9` — every item from that round was addressed in `6541b0d`, as the second review walked through, and a `COMMENTED` review doesn't supersede a changes-requested state.

Head has moved to `52909c0` since I reviewed `e0009a3`, but only by a merge of `main` carrying #380 (defmulti attr-map); this PR's own files are byte-identical to what I reviewed:

```console
$ git diff --stat e0009a3 52909c0 -- crates/cljrs/src/native/pinned.rs crates/cljrs-project/ \
    crates/cljrs/tests/pinned_dylib_e2e.rs crates/cljrs/README.md docs/book/src/cli/deps.md \
    crates/cljrs/src/session.rs crates/cljrs/src/commands/deps.rs
(no output)
```

`cargo fmt --check` is clean at this head. CI's "Build, test, lint" is still running on the merge — worth a glance before merging, though it's the unrelated `cljrs-runtime` change being exercised.

The four remaining items from the second review are now #388, so they're tracked rather than gating: a working-tree digest and cargo's mtime fingerprint can disagree and publish a stale library, `publish_artifact`'s non-atomic copy can poison the cache permanently, the widened digest walk follows symlinks and reads the whole root, plus two smaller ones. None is on the happy path.

---

PR #380, opened 2026-09-08
defmulti: accept the attr-map arity

BuddhiLW:
Clojure's signature is `(defmulti name docstring? attr-map? dispatch-fn & options)`. Only the docstring was recognised, so with an attribute map present the **map itself became the dispatch function**.

Nothing failed at definition time, because a map is callable. Every call dispatched through `{:added "1.0"}`, which returns nil for any argument, so the error surfaced later and somewhere else:

```clojure
(defmulti area "Area of a shape." {:added "1.0"} :kind)
(defmethod area :square [s] (* (:side s) (:side s)))
(area {:kind :square :side 4})
;; => No method in multimethod 'area' for dispatch value nil
```

The message points at the call, and the call is correct — the definition four lines up is where the meaning was lost.

### The change

The head is parsed as a unit. Each optional part is taken as such only when a form still follows it, so a map that genuinely IS the dispatch function is left alone:

```clojure
(defmulti lookup {:a :first})   ; still dispatches through the map
```

Options are located relative to the dispatch fn, whose index moves with whatever precedes it — `:default` was previously found at a fixed offset.

The docstring also reaches the var's metadata now, alongside the attr map. Previously it was skipped for parsing purposes and then discarded.

### Tests

Eight cases: all four head shapes dispatching correctly, the regression stated directly, docstring and attr-map metadata, `:default` after a full head, a map still usable as the dispatch fn, and the arity error when no dispatch fn is present. `cargo test -p cljrs-runtime` is green.

### DIFF
--- a/crates/cljrs-runtime/src/interp/special.rs
+++ b/crates/cljrs-runtime/src/interp/special.rs
@@ -2272,7 +2272,7 @@ fn parse_defmulti_head<'a>(args: &'a [Form]) -> EvalResult<DefmultiHead<'a>> {
     let mut docstring = None;
     let mut attr_map = None;
 
-    if i + 1 < args.len()
+    if i < args.len()
         && let Some(s) = args[i].as_string()
     {
         docstring = Some(s.to_string());
@@ -2314,13 +2314,15 @@ fn eval_defmulti(args: &[Form], env: &mut Env) -> EvalResult {
         i += 2;
     }
 
-    // Metadata sources, weakest first: the attr map, then `^` marks on the
-    // name, then the docstring.
+    // Metadata sources, weakest first: `^` marks on the name, then the attr
+    // map, then the docstring.  Clojure builds the same order in
+    // `clojure.core/defmulti` -- `(conj (meta mm-name) m)` puts the attr map,
+    // docstring already merged into it, on top of the name's metadata.
     let attr_meta = match head.attr_map {
         Some(form) => Some(eval(form, env)?),
         None => None,
     };
-    let meta = merge_meta(attr_meta, name_meta);
+    let meta = merge_meta(name_meta, attr_meta);
     let meta = merge_meta(meta, head.docstring.as_deref().map(doc_meta));
 
     let mfn = MultiFn::new(name_arc.clone(), dispatch_fn, default_dispatch);
--- a/crates/cljrs-runtime/src/interp/special.rs
+++ b/crates/cljrs-runtime/src/interp/special.rs
@@ -2253,28 +2253,57 @@ fn build_impl_fn(
 
 // ── defmulti ──────────────────────────────────────────────────────────────────
 
-fn eval_defmulti(args: &[Form], env: &mut Env) -> EvalResult {
-    // (defmulti name dispatch-fn-form) or (defmulti name "doc" dispatch-fn :default val)
-    let (name, name_meta) = require_sym_meta(args, 0, "defmulti", env)?;
-    let name_arc: Arc<str> = Arc::from(name.as_str());
+/// The parsed head of a `defmulti` form.
+///
+/// `(defmulti name docstring? attr-map? dispatch-fn & options)` — the two
+/// optional parts sit between the name and the dispatch function, and each is
+/// only taken as such when a form still follows it, so the dispatch function
+/// is never mistaken for one of them.
+struct DefmultiHead<'a> {
+    docstring: Option<String>,
+    attr_map: Option<&'a Form>,
+    /// Index of the dispatch-fn form; `args[dispatch_idx + 1..]` are options.
+    dispatch_idx: usize,
+}
 
-    let rest_start = if args.len() > 2 && args[1].as_string().is_some() {
-        2
-    } else {
-        1
-    };
+/// Split a `defmulti`'s arguments into its head parts. Pure: reads shapes only.
+fn parse_defmulti_head<'a>(args: &'a [Form]) -> EvalResult<DefmultiHead<'a>> {
+    let mut i = 1;
+    let mut docstring = None;
+    let mut attr_map = None;
 
-    if args.len() <= rest_start {
+    if i + 1 < args.len()
+        && let Some(s) = args[i].as_string()
+    {
+        docstring = Some(s.to_string());
+        i += 1;
+    }
+    if i + 1 < args.len() && matches!(args[i].kind, FormKind::Map(_)) {
+        attr_map = Some(&args[i]);
+        i += 1;
+    }
+    if i >= args.len() {
         return Err(EvalError::Runtime(
             "defmulti requires a dispatch function".into(),
         ));
     }
+    Ok(DefmultiHead {
+        docstring,
+        attr_map,
+        dispatch_idx: i,
+    })
+}
 
-    let dispatch_fn = eval(&args[rest_start], env)?;
+fn eval_defmulti(args: &[Form], env: &mut Env) -> EvalResult {
+    let (name, name_meta) = require_sym_meta(args, 0, "defmulti", env)?;
+    let name_arc: Arc<str> = Arc::from(name.as_str());
+    let head = parse_defmulti_head(args)?;
+
+    let dispatch_fn = eval(&args[head.dispatch_idx], env)?;
 
     // Parse optional :default val.
     let mut default_dispatch = ":default".to_string();
-    let mut i = rest_start + 1;
+    let mut i = head.dispatch_idx + 1;
     while i + 1 < args.len() {
         if let FormKind::Keyword(k) = &args[i].kind
             && k == "default"
@@ -2285,11 +2314,20 @@ fn eval_defmulti(args: &[Form], env: &mut Env) -> EvalResult {
         i += 2;
     }
 
+    // Metadata sources, weakest first: the attr map, then `^` marks on the
+    // name, then the docstring.
+    let attr_meta = match head.attr_map {
+        Some(form) => Some(eval(form, env)?),
+        None => None,
+    };
+    let meta = merge_meta(attr_meta, name_meta);
+    let meta = merge_meta(meta, head.docstring.as_deref().map(doc_meta));
+
     let mfn = MultiFn::new(name_arc.clone(), dispatch_fn, default_dispatch);
     let var = env
         .globals
         .intern(&env.current_ns, name_arc, Value::MultiFn(GcPtr::new(mfn)));
-    if let Some(meta_val) = name_meta {
+    if let Some(meta_val) = meta {
         var.get().set_meta(meta_val);
     }
     Ok(Value::Var(var))
--- a/crates/cljrs-runtime/tests/defmulti_attr_map.rs
+++ b/crates/cljrs-runtime/tests/defmulti_attr_map.rs
@@ -1,0 +1,135 @@
+//! `defmulti`'s optional docstring and attribute map.
+//!
+//! Clojure's signature is `(defmulti name docstring? attr-map? dispatch-fn &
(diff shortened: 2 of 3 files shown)


crates/cljrs-runtime/tests/defmulti_attr_map.rs
```diff
@@ -0,0 +1,135 @@
...
+
+/// Evaluate every form in `src` in one fresh environment; yield the last value.
+fn eval_fresh(src: &str) -> Result<Value, String> {
+    let (_, mut env) = make_env();
+    let mut parser = Parser::new(src.to_string(), "<test>".to_string());
+    let forms = parser.parse_all().map_err(|e| format!("parse: {e:?}"))?;
+    let mut result = Value::Nil;
+    for form in forms {
+        result =
+            cljrs_runtime::interp::eval::eval(&form, &mut env).map_err(|e| format!("eval: {e:?}"))?;
```
csm: `cargo fmt --check` fails here, which is what turns the `Build, test, lint` job red — it is the only failure in that job. rustfmt wants:

```rust
        result = cljrs_runtime::interp::eval::eval(&form, &mut env)
            .map_err(|e| format!("eval: {e:?}"))?;
```

---

crates/cljrs-runtime/src/interp/special.rs
```diff
@@ -2253,28 +2253,57 @@ fn build_impl_fn(
...
+fn parse_defmulti_head<'a>(args: &'a [Form]) -> EvalResult<DefmultiHead<'a>> {
+    let mut i = 1;
+    let mut docstring = None;
+    let mut attr_map = None;
+
+    if i + 1 < args.len()
+        && let Some(s) = args[i].as_string()
+    {
+        docstring = Some(s.to_string());
+        i += 1;
```
csm: The `i + 1 < args.len()` lookahead earns its keep on the map branch below — a map is callable, so `(defmulti lookup {:a :first})` has a sensible reading as a dispatch fn. It costs on the string branch, though: a string is never callable, so the lookahead only defers an error the PR could catch at definition time.

```clojure
(defmulti area "Area.")
(defmethod area :square [_] 1)
(area {:kind :square})
```

The string becomes the dispatch fn and this fails at the call with `NotCallable("<string> is not callable")` — the same shape of late, misdirected failure the PR is fixing. Dropping the guard on this branch only (`if let Some(s) = args[i].as_string()`) makes it `defmulti requires a dispatch function`, raised at the definition, and also matches Clojure, which takes a leading string as the docstring unconditionally.

---

crates/cljrs-runtime/src/interp/special.rs
```diff
@@ -2285,11 +2314,20 @@ fn eval_defmulti(args: &[Form], env: &mut Env) -> EvalResult {
...
     }
 
+    // Metadata sources, weakest first: the attr map, then `^` marks on the
+    // name, then the docstring.
+    let attr_meta = match head.attr_map {
+        Some(form) => Some(eval(form, env)?),
+        None => None,
+    };
+    let meta = merge_meta(attr_meta, name_meta);
+    let meta = merge_meta(meta, head.docstring.as_deref().map(doc_meta));
```
csm: The attr map is the weakest source here; in Clojure it is the strongest. `clojure.core/defmulti` builds its metadata as

```clojure
m (if docstring (assoc m :doc docstring) m)
m (if (meta mm-name) (conj (meta mm-name) m) m)
```

`conj` onto the name's metadata puts `m` — the attr map, docstring already merged in — on top, so the attr map wins a key collision with a `^` mark. This PR inverts that:

```clojure
(defmulti ^{:added "name"} area {:added "attr"} :kind)
(:added (meta (var area)))
;; => "name"   ; Clojure: "attr"
```

Docstring-over-attr-map (`:doc` in the attr map losing to the docstring argument) already matches Clojure, so only the `merge_meta(attr_meta, name_meta)` order needs flipping. `the_attr_map_lands_in_the_var_metadata` uses disjoint keys, so it passes either way — a collision case would pin the order down.

---

csm (commented):
The diagnosis is right and the fix is in the right place. I checked that `eval_defmulti` is the only site that parses a `defmulti` head: `cljrs-ir`'s ANF lowerer rejects `defmulti` as interpreter-only (`lower/anf.rs:736`) and AOT routes it to the interpreter (`aot.rs:1363`), so there is no second parser to keep in sync. Locally: the 8 new tests pass, `cargo clippy -p cljrs-runtime --all-targets` is clean.

Three things to fix, then two notes.

**1. CI is red on formatting only.** `cargo fmt --check` fails on `tests/defmulti_attr_map.rs:33`. That is the whole of the `Build, test, lint` failure — inline comment has the replacement.

**2. Attr-map metadata should beat `^` metadata on the name, not lose to it.** Clojure's `(conj (meta mm-name) m)` puts the attr map on top. Inline comment with the diverging case.

**3. The lookahead on the docstring branch defers an error it could catch.** `(defmulti area "Area.")` makes the string the dispatch fn and fails at the call site. Inline comment.

**Notes, not blockers:**

Redefining a `defmulti` no longer clears the var's metadata, and this PR is the first thing to put anything interesting there:

```clojure
(defmulti area "D" {:added "1.0"} :kind)
(defmulti area :kind)
[(:doc (meta (var area))) (:added (meta (var area)))]
;; => ["D" "1.0"]
```

`set_meta` is only called when the new metadata is `Some`, so the old map survives. `eval_def` has the same shape, so it is a pre-existing quirk rather than something this PR introduces — but before this change a `defmulti` docstring was discarded outright, so nothing could get stranded. Clojure re-runs `(def ~(with-meta mm-name m))` on every `defmulti` and replaces the metadata. (Relatedly, and squarely pre-existing: re-evaluating a `defmulti` here interns a fresh `MultiFn` and drops the method table, where Clojure's is idempotent and keeps it.)

Per `CLAUDE.md`, `crates/cljrs-runtime/README.md` enumerates every file under `tests/`; `defmulti_attr_map.rs` should be added to that listing in this commit.

Out of scope, but adjacent enough to be worth recording somewhere: `:hierarchy` is still ignored, and unknown options pass silently — `(defmulti area :kind :bogus 1)` is accepted, where Clojure's `check-valid-options` throws.

---

BuddhiLW:
All three fixed in `f10875d`.

**1. Formatting / red CI.** Applied your diff verbatim. `cargo fmt --all --check` is clean.

**2. Attr map beats `^` on the name.** You are right, and it is worth naming why the existing test missed it: `the_attr_map_lands_in_the_var_metadata` uses disjoint keys, so precedence never came up. Flipped to `merge_meta(name_meta, attr_meta)` — docstring-over-attr-map already matched Clojure and is untouched. Two collision tests now pin the whole order:

```clojure
(defmulti ^{:added "name"} area {:added "attr"} :kind)
(:added (meta (var area)))                     ;; => "attr"

(defmulti area "from the docstring" {:doc "from the attr map"} :kind)
(:doc (meta (var area)))                       ;; => "from the docstring"
```

Mutation-checked: restoring the old order fails `the_attr_map_beats_metadata_on_the_name` and nothing else.

**3. The lookahead on the string branch.** Dropped, exactly as you framed it — the map branch keeps it because a map genuinely is callable, the string branch does not because a string never is. `(defmulti area "Area.")` now raises `defmulti requires a dispatch function` at the definition instead of `NotCallable` at the first call. Also mutation-checked.

**README.** `defmulti_attr_map.rs` added to the `tests/` listing in the same commit.

### On your notes

**Metadata surviving a redefinition** — reproduced, and it is as you diagnosed: `set_meta` only runs when the new metadata is `Some`, so `(defmulti area "D" {:added "1.0"} :kind)` followed by a bare `(defmulti area :kind)` keeps both keys. Since `eval_def` has the same shape, fixing it here would leave `def` diverging from `defmulti` for no reason — the fix belongs at whatever both call, not at one of them. Left alone deliberately rather than overlooked; happy to take it as its own change together with the non-idempotent `MultiFn` re-intern you mention, which has the same root.

**`:hierarchy` ignored, unknown options silent** — filed on our side. `(defmulti area :kind :bogus 1)` being accepted where Clojure's `check-valid-options` throws is the same shape of late/absent failure this PR is about, so it wants a real fix rather than a note; out of scope here.

`cargo test -p cljrs-runtime` green (52 binaries), `cargo clippy -p cljrs-runtime --all-targets` clean.

crates/cljrs-runtime/src/interp/special.rs
```diff
@@ -2253,28 +2253,57 @@ fn build_impl_fn(
...
-    } else {
-        1
-    };
-
-    if args.len() <= rest_start {
+/// The parsed head of a `defmulti` form.
+///
+/// `(defmulti name docstring? attr-map? dispatch-fn & options)` — the two
+/// optional parts sit between the name and the dispatch function, and each is
+/// only taken as such when a form still follows it, so the dispatch function
```
csm, line 2260: This paragraph describes the previous rule. Dropping the lookahead on the string branch means the two optional parts no longer share one condition: the docstring is taken whenever a leading string is present, and only the attr map still requires a form behind it. Something like:

> A leading string is always the docstring — a string is never callable, so there is nothing to lose by claiming it, and a `defmulti` left without a dispatch function is rejected here rather than at its first call. A map is callable, so one is taken as the attr map only when a form still follows it: `(defmulti lookup {:a :first})` dispatches through the map.

That also puts the asymmetry's justification next to the code that relies on it.

---

csm (commented):
All three points from the last round are fixed, and `f10875d` is a clean piece of work. Verified against the merge head `f05c041`:

- Metadata precedence now matches Clojure. `(defmulti ^{:added "n"} area {:added "a"} :kind)` gives `"a"`; docstring still beats `:doc` in the attr map; a name-meta key the attr map does not carry survives (`^{:private true}` with `{:added "a"}` gives `[true "a"]`).
- `(defmulti area "Area.")` is now `defmulti requires a dispatch function` at the definition, instead of `NotCallable` at the first call.
- `cargo fmt --check` is clean, `cargo clippy -p cljrs-runtime --all-targets` is clean, and all 11 tests in `defmulti_attr_map.rs` pass. The two new precedence tests pin the order on colliding keys, which is what the old ones could not do.

One thing to restore, and it is not from your commits:

**The merge commit `f05c041` dropped the README entry.** `f10875d` added a `defmulti_attr_map.rs` block to `crates/cljrs-runtime/README.md` under `tests/`. `main` had meanwhile added a `defmethod_cross_ns.rs` block in the same place, and the conflict was resolved by taking `main`'s side, so the `defmulti_attr_map.rs` lines are gone from the head. Both belong there — `git show f10875dc:crates/cljrs-runtime/README.md` has the text to put back.

The inline comment is a stale doc comment on `DefmultiHead`, left behind by the string-branch fix.

Two notes, neither blocking and both pre-existing in shape:

`(defmulti area "Doc." {:added "1.0"})` is the one head the map lookahead still reads differently from Clojure. The docstring is claimed unconditionally now, so the map has nothing behind it and becomes the dispatch fn: `:doc` lands on the var, `:added` does not, and a call fails with `No method in multimethod 'area' for dispatch value nil`. It follows from the rule the PR documents — a trailing map is a dispatch fn — and Clojure's reading of that form (nil dispatch fn) is broken too, just differently. Worth knowing about rather than changing.

Redefining still leaves stale metadata behind, since `set_meta` runs only when the new metadata is `Some`:

```clojure
(defmulti area "D" {:added "1.0"} :kind)
(defmulti area :kind)
[(:doc (meta (var area))) (:added (meta (var area)))]
;; => ["D" "1.0"]
```

`eval_def` has the same shape, so it is not this PR's to fix — but this PR is what makes a `defmulti` var carry anything worth stranding.

---

PR #381, opened 2026-09-08
One namespace-resolution rule, not seven copies

BuddhiLW:
Follow-up to the note on #378 ("the three-line alias resolution is now in its sixth copy").

"Resolve a symbol's namespace part: a `:require :as` alias wins, anything else is taken literally, and an absent one means the current namespace" was open-coded at **seven** sites — `eval_symbol`, the `var` form, macro resolution, syntax-quote, `binding`, protocol names, and `eval_special`'s qualified lookup.

They agreed. But nothing *made* them agree, so the agreement was a property of whoever last remembered to check, and a construct added later starts from whichever copy got pasted.

### Why this is worth a PR rather than a cleanup

Extending the rule is exactly what has been going wrong. `eval_symbol` grew a privacy check and versioned-symbol routing; the other six copies did not. So today every construct that reaches into another namespace enforces a different subset of the same rule — which is how #378 needed two follow-up guards for behaviour `eval_symbol` already had, and why it would have been reasonable to assume it didn't.

`Env::resolve_ns_part` and `Env::resolve_ns_or_current` name the rule once. The next thing it has to learn is learned by all of them.

### Scope

No behaviour changes — the extracted body is the copies' body, and `cargo test -p cljrs-runtime` is green, along with fmt and clippy.

`defmethod` is deliberately **not** included: its cross-namespace resolution lands in #378, so it becomes the eighth caller there rather than here. `resolve_protocol_sym`'s `None` arm is left alone too — it looks the whole symbol up rather than falling back to the current namespace, which is a genuinely different rule and not a copy of this one.

### The test

`ns_part_resolution.rs` is the part I'd argue for keeping even if the refactor were rejected. It evaluates the same expression twice — through an alias and through the full namespace name — for every construct that resolves one, and requires the two to agree:

```clojure
{ns}/value                                  ; symbol
@(var {ns}/value)                           ; var form
({ns}/twice 21)                             ; macro call
(str `{ns}/value)                           ; syntax-quote
(binding [{ns}/*knob* :bound] {ns}/*knob*)  ; binding target
(extend-protocol {ns}/Shaped …)             ; protocol name
```

with `{ns}` as `l` and then `some.lib`. Before, each of those was covered separately if at all; the agreement between them was covered nowhere. A construct that stops going through the seam now fails a test instead of quietly diverging.

---

Happy to fold this into #378 instead if you'd rather review one diff — I kept it separate because it's a pure refactor and #378 is a fix.

### DIFF
--- a/crates/cljrs-compiler/src/aot.rs
+++ b/crates/cljrs-compiler/src/aot.rs
@@ -1468,10 +1468,7 @@ fn pin_versioned_references(
     for (ns_part, commit) in pins {
         let base: Arc<str> = match &ns_part {
             Some(p) => {
-                let resolved = env
-                    .globals
-                    .resolve_alias(&env.current_ns, p)
-                    .unwrap_or_else(|| Arc::from(p.as_str()));
+                let resolved = env.globals.resolve_ns_part_in(&env.current_ns, p);
                 Arc::from(cljrs_runtime::env::versioned::base_ns_name(&resolved))
             }
             None => Arc::from(cljrs_runtime::env::versioned::base_ns_name(&env.current_ns)),
--- a/crates/cljrs-nrepl/src/engine.rs
+++ b/crates/cljrs-nrepl/src/engine.rs
@@ -386,10 +386,7 @@ impl Engine {
 
         if let Some((alias, name_prefix)) = prefix.split_once('/') {
             // Qualified prefix: complete interns of the aliased/named namespace.
-            let full = self
-                .globals
-                .resolve_alias(&context_ns, alias)
-                .unwrap_or_else(|| Arc::from(alias));
+            let full = self.globals.resolve_ns_part_in(&context_ns, alias);
             if let Some(ns) = namespaces.get(&full) {
                 for (name, var) in ns.get().interns.lock().unwrap().iter() {
                     if name.starts_with(name_prefix) {
@@ -461,10 +458,7 @@ impl Engine {
         let sym = req.sym.clone().unwrap_or_default();
         let var = match sym.split_once('/') {
             Some((ns_part, name)) => {
-                let full = self
-                    .globals
-                    .resolve_alias(&context_ns, ns_part)
-                    .unwrap_or_else(|| Arc::from(ns_part));
+                let full = self.globals.resolve_ns_part_in(&context_ns, ns_part);
                 self.globals.lookup_var_in_ns(&full, name)
             }
             None => self.globals.lookup_var_in_ns(&context_ns, &sym),
--- a/crates/cljrs-runtime/README.md
+++ b/crates/cljrs-runtime/README.md
@@ -118,10 +118,12 @@ tests/
                                      protocol method body; params shadow them
   deftype_types.rs                 — deftype: positional ctor, `.-field`, protocol
                                      impls, and mutable fields written with `set!`
-  ns_part_resolution.rs            — one namespace-resolution rule: an alias and
+  ns_part_resolution.rs            — one alias-resolution rule: an alias and
                                      a full namespace name agree for symbols,
                                      `var`, macros, syntax-quote, `binding`
-                                     and protocol names
+                                     and protocol names, and an absent ns part
+                                     means the current namespace (each test
+                                     fails when the arm it names is stubbed)
   qualified_protocol_impl.rs       — a qualified protocol name in an impl position
                                      (defrecord/deftype/reify/extend-*) resolves
                                      through its own namespace
@@ -309,6 +311,19 @@ pub fn ir_cache(&self) -> &Arc<tiered::ir_cache::IrCache>;
 pub fn eval(&self, form: &Form, env: &mut Env) -> EvalResult;
 pub fn call_cljrs_fn(&self, f: &CljxFn, args: &[Value], env: &mut Env) -> EvalResult;
 pub fn on_fn_defined(&self, f: &CljxFn, env: &mut Env);
+
+/// The namespace a qualified symbol's `ns` part names: a `:require … :as`
+/// alias wins, anything else is taken literally.  This is the ONE definition
+/// of that rule; every construct that resolves a qualified name reads it.
+/// `Env::resolve_ns_part` is the same rule relative to that env's
+/// `current_ns`, and `Env::resolve_ns_or_current` adds the absent-ns-part
+/// case.  Callers relative to something else — `resolve`, which follows the
+/// `*ns*` dynamic var, and versioned resolution, which follows a defining
+/// namespace — name it here instead of open-coding the lookup.
+///
+/// Note what this is NOT: `eval_symbol`'s privacy check and versioned-symbol
+/// routing sit *after* its call to this, and remain its own.
+pub fn resolve_ns_part_in(&self, current_ns: &str, ns_part: &str) -> Arc<str>;
 
 /// Copy `src_ns`'s interns into `dst_ns` as refers.  Both are *explicit*
 /// refers (`(:require [x :refer :all])` / `:refer [...]`) and are never
--- a/crates/cljrs-runtime/README.md
+++ b/crates/cljrs-runtime/README.md
@@ -118,6 +118,10 @@ tests/
                                      protocol method body; params shadow them
   deftype_types.rs                 — deftype: positional ctor, `.-field`, protocol
                                      impls, and mutable fields written with `set!`
+  ns_part_resolution.rs            — one namespace-resolution rule: an alias and
+                                     a full namespace name agree for symbols,
+                                     `var`, macros, syntax-quote, `binding`
+                                     and protocol names
   qualified_protocol_impl.rs       — a qualified protocol name in an impl position
                                      (defrecord/deftype/reify/extend-*) resolves
                                      through its own namespace
--- a/crates/cljrs-runtime/src/env/env.rs
+++ b/crates/cljrs-runtime/src/env/env.rs
@@ -313,6 +313,20 @@ impl GlobalEnv {
         aliases.get(alias).cloned()
     }
 
+    /// The namespace a qualified symbol's `ns` part names, read from
+    /// `current_ns`'s alias table.
+    ///
+    /// A `:require … :as` alias wins; anything else is taken literally. This
+    /// is the whole rule, and the only place it is written: callers that
+    /// resolve relative to something other than an `Env`'s `current_ns` —
+    /// `resolve`, which is relative to the `*ns*` dynamic var, and versioned
+    /// resolution, which is relative to a defining namespace — name that
+    /// namespace here rather than open-coding the lookup.
+    pub fn resolve_ns_part_in(&self, current_ns: &str, ns_part: &str) -> Arc<str> {
+        self.resolve_alias(current_ns, ns_part)
+            .unwrap_or_else(|| Arc::from(ns_part))
+    }
+
     /// Resolve an auto-resolved keyword name (the text after `::`) to its
     /// fully-qualified `ns/name` form.
     ///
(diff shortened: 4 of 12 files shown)


csm (commented):
Reviewed at `edce713`. Built and ran locally: `cargo test -p cljrs-runtime` all green (including the 8 new tests), `cargo fmt --all --check` clean, `cargo clippy -p cljrs-runtime --all-targets` clean. The extraction is faithful at all seven sites — each replaced block is the same logic, and `eval_symbol`'s `Arc::from(ns_part.as_ref())` and the helper's `Arc::from(ns_part)` are equivalent. No behaviour change, as claimed.

Three things worth acting on, then nits.

### 1. `an_unqualified_name_means_the_current_namespace` does not reach the seam

`(do (def here 7) here)` never calls `resolve_ns_or_current`. `eval_symbol` returns from `env.globals.lookup_in_ns(&env.current_ns, s)` (eval.rs:251) before it gets to the qualified-symbol branch, so the bare symbol is resolved by a different code path than the one the test is named for.

Mutation check — replacing the `None` arm with `Arc::from("MUTANT.ns")`:

```
test result: ok. 8 passed; 0 failed   # tests/ns_part_resolution.rs
test result: FAILED. 227 passed; 11 failed   # --lib
```

All eight tests in the new file stay green. The existing lib unit tests catch it, but the file whose stated job is pinning this rule does not. A construct that actually goes through the seam unqualified would fix it — `@(var here)`, or an unqualified macro call.

The other six tests do bite: stubbing `resolve_ns_part` to `Arc::from(ns_part)` fails all six agreement tests. That half of the file works as advertised.

### 2. The seam is the part that already agreed, not the part that diverged

The body's argument is that `eval_symbol` grew a privacy check and versioned-symbol routing while the other six copies didn't. Both of those live *after* the `resolve_alias` call in `eval_symbol` (eval.rs:262-281) and are untouched here, so they still don't exist at the other six sites. After this PR:

- `mylib/x` inside `mylib@hash`'s own source routes to the pinned commit; `(var mylib/x)` in the same source does not.
- `eval_symbol` raises `var: other/x is not public`; `(binding [other/*x* v] …)`, `(var other/x)` and a macro call through `other/` do not.

What `resolve_ns_part` unifies is the three-line alias lookup that was already identical everywhere. That's still worth landing — it's the anchor point — but as written the helper can't grow the privacy check (it needs the name and the var) and growing the version routing into it would change behaviour at six sites, i.e. not this PR. Suggest narrowing the claim to what the seam does, and naming the follow-up that moves the divergent parts behind it.

### 3. There's an eighth copy, and it's already diverged

`handle_resolve` in `crates/cljrs-runtime/src/interp/apply.rs:1884` open-codes the same three lines, but against `resolve_current_ns(env)` — the `*ns*` dynamic var — rather than `env.current_ns`. So `(resolve 'alias/x)` and `alias/x` can resolve aliases relative to different namespaces. `resolve_versioned_value` (env/versioned.rs:64) is a ninth, against a `defining_ns` parameter.

Neither fits `resolve_ns_part`'s signature, which is exactly why they went unnoticed in the "seven copies" count. A `resolve_ns_part_in(&self, current_ns: &str, ns_part: &str)` that `resolve_ns_part` delegates to would cover both without changing behaviour; failing that, a line in the doc comment saying the helper is `current_ns`-relative and these sites are deliberately not it.

### Nits

- `LIB` defines `(defmulti describe :kind)` and `(defmethod describe :default …)` that no test uses. Given defmethod is deferred to #378 on purpose, either drop them or say in a comment that they're the fixture #378 will pick up.
- The module doc says the rule must hold for "a `defmethod` target" — true as intent, but nothing in the file tests it, and the doc reads as this file's coverage list.
- `an_unknown_namespace_part_is_taken_literally`'s comment says "whether or not it is loaded yet", but the test `(require 'some.lib)` first, so the not-yet-loaded case isn't covered.
- `env_with_sources` returns an `Arc<GlobalEnv>` that all three callers bind to `_g`. Drop it from the tuple.

On folding into #378: keeping it separate reads right — the refactor is reviewable on its own and #378 then adds the eighth caller rather than the eighth copy.

---

BuddhiLW:
All three addressed in `0fd5e9c`. Point 3 turned out to be bigger than an eighth copy.

### 3. The eighth copy — and the ninth, tenth, eleventh

You are right that neither `handle_resolve` nor `resolve_versioned_value` fits `resolve_ns_part`'s signature, and that this is exactly why they went unnoticed. Grepping for the *shape* rather than for env-relative callers finds four more of the same `resolve_alias(..).unwrap_or_else(|| literal)`:

```
interp/apply.rs:1884      handle_resolve            relative to *ns*
env/versioned.rs:64       resolve_versioned_value   relative to defining_ns
tiered/ir_interp.rs:539   Inst::LoadVar             relative to the lowering ns
tiered/ir_interp.rs:845   versioned var lookup      relative to defining_ns
cljrs-compiler/aot.rs:1471  AOT pin resolution      relative to env.current_ns
cljrs-nrepl/engine.rs:391   completion              relative to the session ns
cljrs-nrepl/engine.rs:466   info/lookup             relative to the session ns
```

So eleven, not seven — and the two you named are the reason: once "which namespace do I read the alias table of" is a *parameter* rather than an assumption, they all fit. Your suggested signature, verbatim:

```rust
pub fn resolve_ns_part_in(&self, current_ns: &str, ns_part: &str) -> Arc<str>
```

on `GlobalEnv`, with `Env::resolve_ns_part` delegating. Env-relative callers are untouched; the other six name the namespace they mean. Still no behaviour change — every one of them was already alias-or-literal.

I deliberately left `rt_abi.rs:2500` and `:2618` alone. They are a fallback *chain* (try the namespace, then the alias, then refers), not a lookup with a literal default — the same function, a different rule. Folding them in would have been the wrong kind of tidy.

### 1. The test that did not reach the seam

Confirmed, and your mutation is the right way to have caught it. `(do (def here 7) here)` returns from `lookup_in_ns(&env.current_ns, s)` at eval.rs:251, before the qualified branch. Changed to `@(var here)`, which has no short-circuit. Re-running your mutation:

```
None => Arc::from("MUTANT.ns")
  an_unqualified_name_means_the_current_namespace ... FAILED    (was: passed)

resolve_ns_part_in => Arc::from(ns_part)
  6 failed  — all six agreement tests, as you measured
  an_unknown_namespace_part_is_taken_literally ... ok           — correct, that IS the literal case
```

I also wrote an unqualified *macro-call* test alongside it and then deleted it: it passed under both mutants, so it pinned nothing. Same criticism you raised, and it seemed worse to ship a second test whose name overstates it than to have one fewer.

### 2. The claim was wider than the seam

Agreed, and narrowed. What is shared is the alias lookup. The privacy check and versioned-symbol routing sit *after* `eval_symbol`'s call to it (eval.rs:262-281) and still exist only there — you are right that the helper cannot grow the privacy check as written (it has neither the name nor the var), and that growing version routing into it would change behaviour at every other site. The module doc and the commit body now say what the seam does and name the two divergences as separate work rather than implying this PR closes them.

Your framing of the follow-up is the right one: `resolve_ns_part` collapses the alias lookup, and the rest of the answer is fewer special forms hand-written in Rust. #382 is that direction for the datatype family.

### Nits

All taken: the unused `defmulti`/`defmethod` dropped from `LIB`; the module doc no longer lists `defmethod` as covered; "whether or not it is loaded yet" removed from a test that requires the namespace first; `env_with_sources` no longer returns the `Arc<GlobalEnv>` all three callers discarded.

`cargo test --workspace` green (161 binaries, 0 failures), `cargo fmt --all --check` clean, `cargo clippy --all-targets` introduces no new warnings (the one that remains is the pre-existing large-`Err` closure in `execution_tier_parity.rs`).

csm (commented):
Re-reviewed at `44cb9f5`. All three points are addressed, and both mutations reproduce exactly as you reported:

```
resolve_ns_part_in => Arc::from(ns_part)
  6 failed (the six agreement tests); an_unknown_namespace_part_is_taken_literally ok

None => Arc::from("MUTANT.ns")
  an_unqualified_name_means_the_current_namespace ... FAILED
```

`cargo test --workspace` green here too (165 binaries, 1514 tests, 0 failures), `cargo fmt --all --check` clean, `cargo clippy -p cljrs-runtime --all-targets` clean. CI is green on this head.

Moving the rule to `GlobalEnv` with the namespace as a parameter is the right call, and I checked each of the seven new sites: all were `resolve_alias(x, y).unwrap_or_else(|| literal)`, so no behaviour change. Your `rt_abi` exclusion is correct — 2500 and 2618 are literal-namespace-first, then alias, then refers, which is a different rule, not a defaulted lookup.

### The merge brought the copy back

`44cb9f5` merges main, which now carries #378. `eval_defmethod` at `crates/cljrs-runtime/src/interp/special.rs:2363` is `resolve_ns_or_current`, open-coded:

```rust
let owner_ns: Arc<str> = match parsed.namespace.as_deref() {
    Some(ns_part) => env
        .globals
        .resolve_alias(&env.current_ns, ns_part)
        .unwrap_or_else(|| Arc::from(ns_part)),
    None => env.current_ns.clone(),
};
```

It is the only `resolve_alias(&env.current_ns, ..)` left in `cljrs-runtime` outside the seam. The scope note says defmethod "becomes the eighth caller there rather than here" — #378 landed and it didn't, and this branch is where the two meet, so here is now the only place it can happen.

I applied the swap locally to check it is as trivial as it looks:

```rust
let owner_ns: Arc<str> = env.resolve_ns_or_current(parsed.namespace.as_deref());
```

`defmethod_cross_ns` (12) and `ns_part_resolution` (8) both pass, clippy clean. Worth doing in this PR: it is the argument the PR is making, demonstrated by the PR's own merge — a construct that landed three days ago started from a pasted copy, and neither test suite noticed. It also makes the README true; as written, "This is the ONE definition of that rule; every construct that resolves a qualified name reads it" is not, while `special.rs:2363` stands.

Adding a `{ns}/describe` case to `agrees()` would then pin it, which is what the module doc originally promised before it was narrowed.

### The count is fourteen

`edce713` converts 7 call sites; `0fd5e9c` converts 7 more — the exact seven your table lists. That is 14, not 11, whether counting call sites or enclosing functions (`eval`, `eval_symbol`, `resolve_macro`, `qualify_symbol`, `eval_var`, `eval_binding`, `resolve_protocol_sym`; `handle_resolve`, `resolve_versioned_value`, `LoadVar`, `load_global_value`, `pin_versioned_references`, `completion`, `info`). Fifteen with defmethod. Worth fixing in the commit body, since the count is the argument.

### Smaller

- Module doc: "The two unqualified cases are the `None` arm" — only `an_unqualified_name_means_the_current_namespace` is. `an_unknown_namespace_part_is_taken_literally` uses a *qualified* name (`some.lib/value`) and exercises the alias arm's literal fallback, which is why it stays green under the alias-arm mutation. Two non-agreement cases, two different arms.
- `Env::resolve_ns_part` and `Env::resolve_ns_or_current` are new public methods and appear in the README only as prose inside `resolve_ns_part_in`'s doc. CLAUDE.md asks for all public functions, and these are what 7 of the 14 sites actually call — a two-line addition to the `env` submodule block.
- `handle_resolve` used to fall back to `ns.clone()` (an `Arc` clone) and now allocates via `Arc::from(ns_part)`. Same value, one allocation on the fully-qualified path. Irrelevant for `resolve`; only worth remembering if `resolve_ns_part_in` ever lands somewhere hot, where an `Option`-returning variant would avoid it.

### One thing that isn't this PR's

`rt_abi` resolves the literal namespace *first* and the alias second — the opposite precedence to the seam. If an alias shadows a real namespace name (`(require '[some.lib :as x])` with a namespace `x` also loaded), compiled code and the interpreter disagree about `x/foo`. Pre-existing, and correctly out of scope here, but it is the same class of divergence and worth an issue so the seam's existence doesn't imply it's settled.

---

PR #382, opened 2026-09-08
The datatype family becomes Clojure macros over a deftype* primitive

BuddhiLW:
`deftype`, `defrecord` and `reify` were three Rust special forms, each carrying its own constructor synthesis, map-constructor synthesis and type-symbol interning. Almost none of that is irreducible.

What genuinely needs the interpreter is minting a type tag and registering method impls with the fields in scope. Everything else is sugar, and Clojure can write it:

```clojure
(defmacro reify [& impls]
  `(make-type-instance (deftype* ~(gensym "reify__") [] ~@impls) {}))
```

So `deftype*` stays in Rust and is the only one of the four still in `SPECIAL_FORMS`. `deftype`, `defrecord` and `reify` move to `bootstrap.cljrs`.

**Net: 195 lines of Rust deleted, 55 lines of Clojure added.**

### What changed in the primitive

`deftype*` now returns the tag it minted, so `reify` feeds it straight to `make-type-instance` rather than mentioning a gensym twice. One dataflow source for the tag is what keeps the two references from desyncing.

### The bug this surfaced

Making `deftype` a macro means its field vector round-trips through macroexpansion — and metadata that arrives that way is **quoted**, because it is already a value and re-analysing it would resolve its contents as code (`value_to_form` does this deliberately).

`meta_form_is_mutable` only understood a bare keyword or map, so `^:unsynchronized-mutable` coming through an expansion was silently dropped. The field became immutable, its value went where no read looks, and the method saw `nil` instead of the constructor argument:

```clojure
(deftype Box [^:unsynchronized-mutable n] Counter (peek-n [_] n))
(peek-n (->Box 7))   ;=> nil
```

Nothing covered it, because every existing test writes `deftype` directly in source where the metadata is unquoted. Fixed by reading through the `Quote` wrapper, with a test that pins a mutable field surviving `deftype` being a macro.

### Verification

`cargo test -p cljrs-runtime -p cljrs` green, fmt and clippy clean. Exercised live over the nREPL as well:

```clojure
[:defrecord 12 :reify 42 :deftype-mutable [1 2]]
```

### Known adjacent gap, not addressed here

A datatype form generated *by a user macro* still fails — `(defmacro defcounter [n] \`(deftype ~n […] Proto (m [_] …)))` reports "No implementation of protocol Counter for type Box", because syntax-quote qualifies the method names in impl position. That predates this PR and is orthogonal to it; I dropped the test I had written for it rather than ship a failing one, and I'm happy to file it separately.


csm (changes requested):
The refactor is sound at the tree-walking level and the `meta_form_is_mutable` Quote fix is correct (it matches what `value_to_form` documents at `crates/cljrs-runtime/src/interp/macros.rs:344-355`). I verified the interpreter semantics that matter and they are preserved:

- `reify` still gets one instance per evaluation with its own captured locals — `(defn mk [n] (reify P (pval [_] n)))`, then `[(pval (mk 1)) (pval (mk 2))]` → `[1 2]`.
- `defrecord` positional + `map->T` constructors, `pr-str` as `#Point{...}`, `instance?`, and a `^:unsynchronized-mutable` `deftype` field all behave as on `main`.

But two lists elsewhere in the tree key off the *names* `deftype` and `reify` in the **macro-expanded** form, and the expansion no longer contains them. Both produce silent or hard failures that CI does not cover.

---

### 1. Blocking — `deftype`/`defrecord`/`reify` inside a function body break under the tiered IR interpreter

`crates/cljrs-ir/src/lower/anf.rs:737` rejects `"defrecord" | "deftype" | "reify"` so a function containing one declines lowering and tree-walks. Lowering runs on the expanded body, so it now sees `deftype*`, which is not in that list — it lowers as a generic call and fails at run time.

```clojure
(defprotocol P (pval [x]))
(defn mk [n] (pval (reify P (pval [_] n))))
(loop [i 0 acc 0] (if (< i 2000) (recur (inc i) (+ acc (mk 1))) acc))
```

- `main`: `2000`
- this PR: `Error: × IR interpreter: var not found user/deftype*`

Same failure with `deftype` (and `defrecord`) written inside the hot function. `reify` inside a function is the idiomatic case, so this is not an edge.

### 2. Blocking — AOT silently miscompiles `deftype` and `reify`

`expanded_needs_interpreter` (`crates/cljrs-compiler/src/aot.rs:1407`) matches `"reify" | "deftype"` on the expanded form to force those top-level forms into the interpreted preamble. After this PR that match never fires, the form is compiled natively, and the program returns `nil` with exit 0 — no error.

```clojure
(defprotocol P (pval [x]))
(deftype Box [n] P (pval [_] n))
(println (pval (->Box 9)))
```

- `main` (`cljrs compile`): `[aot] 2 form(s) will be interpreted at startup` → prints `9`
- this PR: `[aot] 1 form(s) will be interpreted at startup` → prints `nil`

Same for the `reify` equivalent (`main` prints `7`, this PR prints `nil`). `defrecord` survives only by accident: `needs_interpreter` (`aot.rs:1365`) bails on it *before* expansion. This is exactly the corruption the comment above the `anf.rs` rejection list warns about — a generic call resolving to a nil-returning stub.

### Fix

Two lines; I applied and verified them locally:

```diff
--- a/crates/cljrs-ir/src/lower/anf.rs
+++ b/crates/cljrs-ir/src/lower/anf.rs
-        | "defrecord" | "deftype" | "reify" | "defn-" | "." => Err(LowerError::UnsupportedForm(
+        | "defrecord" | "deftype" | "deftype*" | "reify" | "defn-" | "." => Err(LowerError::UnsupportedForm(

--- a/crates/cljrs-compiler/src/aot.rs
+++ b/crates/cljrs-compiler/src/aot.rs
-                    || matches!(s.as_str(), "." | "reify" | "deftype" | "defn-")
+                    || matches!(s.as_str(), "." | "reify" | "deftype" | "deftype*" | "defn-")
```

With those, all four cases above match `main` again (`2000`, `2000`, `9`, `7`). Keeping the old names in the lists is still right — they are what a pre-expansion path sees.

### Test coverage this needs

Nothing in CI caught either. `crates/cljrs-compiler/tests/aot_e2e.rs` covers `defrecord` only, and is behind the `aot_full_test` feature. Two additions would pin both:

- A `reify`-in-a-function test in the style of `crates/cljrs-runtime/tests/deftype_mutable_tiered.rs`, which already flips `force_eager_lowering()` in its own binary — that harness reproduces finding 1 directly and runs in normal CI.
- An `aot_e2e.rs` case for `deftype` and one for `reify`, alongside the existing `test_defrecord`.

---

### Non-blocking

**Stale README.** `crates/cljrs-runtime/README.md:1025-1032` still documents `build_positional_ctor`, `build_map_ctor` and `intern_type_symbol`, all deleted here. CLAUDE.md requires the crate README to be updated in the same commit. If the fix above lands, `crates/cljrs-ir/README.md:156-157` and `crates/cljrs-compiler/README.md:467` also enumerate the interpreter-only forms and would want `deftype*`.

**Dead stubs.** `crates/cljrs-runtime/src/builtins/builtins.rs:1454`, `:1455` and `:1459` register `defrecord`, `reify` and `deftype` as `builtin_stub_nil`. The bootstrap `defmacro`s now overwrite those vars, so the stubs are unreachable; worth deleting, or leaving a note on why they stay.

**Arity error.** `(deftype Bad)` used to report `deftype requires a name and field vector`; it now reports `arity error calling deftype: expected 4+, got 3`, which names neither the real arity nor the real argument count. That is the generic macro-arity reporter rather than anything this PR introduced, but `deftype` is a form users get wrong, so it is a visible downgrade.

**Expansion cost.** `reify` now pays a macroexpansion on every evaluation where it used to be a special form. Registry growth per evaluation is unchanged from `eval_reify`, so no regression there — just noting the added expansion.

---

BuddhiLW:
Both blocking findings confirmed and fixed, but not with the two-line diff — the diff would have been the third copy of the thing that broke.

Three passes decide "only the tree-walking interpreter can run this", each with its own hard-coded list. Two of them run on the *expanded* form; only `anf.rs` had `deftype*`. So the finding is not "`deftype*` is missing from a list", it is "the list exists three times", and both silent failures are what that costs.

`crates/cljrs-ir/src/lower/dispatch_family.rs` now states membership once — surface names and `*` primitives alike — and `anf.rs`, `needs_interpreter` and `expanded_needs_interpreter` all read it. `cljrs-ir` is where it can live: `cljrs-runtime` and `cljrs-compiler` both depend on it and neither depends on the other. Each site keeps only its own extras (`defn-`/`.` at the lowerer; `ns`/`require`/`defmacro`/`defonce` at `needs_interpreter`).

All four of your cases now match `main`:

```
reify in a hot fn (tiered)   2000
deftype in a hot fn          2000
cljrs compile + deftype      [aot] 2 form(s) interpreted -> 9
cljrs compile + reify        [aot] 2 form(s) interpreted -> 7
```

### Tests

Your two suggestions, plus the one that actually pins the invariant:

- **`crates/cljrs-runtime/tests/dispatch_family_expansion.rs`** — macroexpands each surface form and asserts the primitive it expands *through* is a family member. Change what `deftype` expands to and the first assertion fires; forget to add the new primitive and the second does. This is the test whose absence let the two lists drift.
- **`crates/cljrs-runtime/tests/datatype_in_fn_tiered.rs`** — `deftype`/`defrecord`/`reify` inside a function body, driven past the warm threshold under `force_eager_lowering()` in its own binary, per your pointer to `deftype_mutable_tiered.rs`. Verified it fails 4/4 with the lowerer's `deftype*` entry removed.
- **`aot_e2e.rs`** — `test_deftype` and `test_reify` alongside `test_defrecord`. You were right that `defrecord` survived by accident: `needs_interpreter` bails on it before expansion, so it never reached the broken path.

### Non-blocking, all taken

- **Stale README** — `cljrs-runtime/README.md` no longer documents `build_positional_ctor` / `build_map_ctor` / `intern_type_symbol` (those went in a later commit on this branch). `cljrs-ir/README.md` and `cljrs-compiler/README.md` now describe the shared predicate rather than re-listing the forms, so there is no fourth copy in prose either.
- **Dead stubs** — `defrecord`, `reify` and `deftype` `builtin_stub_nil` registrations removed; the bootstrap `defmacro`s bind those vars, and `(resolve 'deftype)` still answers `#'clojure.core/deftype`. I added one for `deftype*`, which had none, so it answers like `protocol*` does.
- **Arity error** — the reporter was counting the implicit `&form`/`&env`. It now subtracts them for macros: `(deftype Bad)` reads `expected 2+, got 1`. Ordinary fn reporting is unchanged, with a test pinning both.
- **Expansion cost** — noted, no change.

`cargo test --workspace` is green (165 binaries, 0 failures); `cargo fmt --all` and `cargo clippy --all-targets` clean.

### One thing to flag about scope

You reviewed `db36a1d`. The branch has since grown three more commits in the same direction, which this fix sits on top of:

```
a9b4a37  extend-type/extend-protocol become macros over `extend`
de6d35b  defprotocol becomes a macro over protocol*
d4617a4  defmulti/defmethod become macros over multi-fn and add-method
c0713c3  one definition of the datatype family, read by every pass
```

So `reify` is a bootstrap macro now too, not the special form you exercised, and `protocol*` is a second `*` primitive — which is why the shared list holds both. Happy to split those into their own PR if you would rather review this one as it stood.

BuddhiLW:
The red check on c0713c3 is not a test failure: the job log ends in "The runner has received a shutdown signal" during `cargo test --workspace`, with clippy and fmt already green. I cannot re-run it from the fork; a re-run of the failed job should go green.

BuddhiLW:
Rebased onto `main`. The conflict was not textual: `main` has since merged #378 and #380, which
added docstring/attr-map parsing and metadata precedence to `eval_defmulti`, and version-pin,
versioned-namespace and privacy refusals to `eval_defmethod`. This branch deletes both of those
Rust functions, so a mechanical "take ours" would have silently reverted two merged PRs.

Four commits carry that forward instead:

- **`defmulti` keeps its attr-map arity.** The macro now parses the same head shape the special
  form did. The two optional parts are decided differently on purpose: a leading string is ALWAYS
  the docstring, while a map is the attr-map only when a form still follows it, because a map is
  itself callable and `(defmulti lookup {:a :first})` is a legitimate map-valued dispatch fn.
  Metadata precedence is preserved as precedence, not merely presence: attr map on top of the
  name's `^` marks, explicit docstring over a `:doc` the attr map carries. It rides on the name
  symbol so it lands on the var `def` binds. `defmulti_attr_map.rs` (11) is green unchanged.

- **`defmethod` refuses a versioned target again.** The pin can arrive in either half of the
  name, and the namespace half is the one that survives `Symbol::parse` (a `(require '[mylib@abc1234 :as v1])`
  alias resolves straight to the versioned namespace, so `parsed.version` is None). Both are
  refused, gated on the namespace boundary so a versioned namespace can still extend its own
  multimethods while it loads.

- **`defmethod`'s two failure modes stay distinguishable.** "not defined" and "bound to a X, not a
  multimethod" are different mistakes and the message says which.

- **The versioned fixture gets the state its loader actually leaves.** This one is the interesting
  regression. `Env::new(globals, "mylib@abc1234")` sets a namespace NAME and nothing else, while
  the real loader creates the namespace, pre-refers `clojure.core` into it, and points `*ns*` at
  it. While `defmulti` was a special form none of that mattered: a special form is ambient in every
  namespace and reads no dynamic var. Hoisting it into a `clojure.core` macro changed WHO CAN SEE
  it, and two tests failed with `UnboundSymbol("defmulti")`. Nothing in the hoist diff mentions
  namespaces, which is exactly why it surfaced as a test failure rather than as a review comment.

Verification on the rebased head: `cargo test --workspace` green (169 suites, 1551 tests), `cargo fmt --check` clean,
`cargo clippy --workspace -- -D warnings` clean.

BuddhiLW:
The rebased head is red on `Build, test, lint` again, and it is the same cause as at `c0713c3`, now diagnosed rather than guessed.

```
2026-09-12T16:10:40Z test interp::eval::tests::prop_splice_agrees_across_eval_quote_and_syntax_quote has been running for over 60 seconds
2026-09-12T16:15:23Z ##[error]The runner has received a shutdown signal.
```

`Check formatting`, `Check` and `Clippy` all passed; `Tests` was **cancelled**, not failed. The runner is running out of memory: three splice properties in `interp/eval.rs` call `eval_str`, which builds a whole runtime per call, and they peak at **17.4 GB RSS / 3m34s** on their own.

That is #389, which is independent of this PR and unblocks CI for every open branch. With it, the same three properties take 0.33 GB / 3.0s, and the `cljrs-runtime` lib suite goes from 135.38s to 1.60s for the same 238 tests.

This branch is green locally at the rebased head: `cargo test --workspace` 169 suites / 1551 tests, `cargo fmt --check` and `cargo clippy --workspace -- -D warnings` clean.

Suggested order: merge #389 first, then re-run this one.

csm (commented):
Re-reviewed at `837567e`. **Both blocking findings from my earlier review are fixed, and I verified each by running them** — treat that changes-requested review as stale.

`dispatch_family.rs` is a better answer than the two-line patch I proposed: membership stated once, read by all three passes, with the namespace qualifier stripped so `clojure.core/deftype` is a member too. `dispatch_family_expansion.rs` pins the expansion and the list against each other, which is the part that would otherwise rot.

Verified on a release build of this head, against a release build of `main` on the same machine:

| case | `main` | before | now |
|---|---|---|---|
| `reify` in a hot fn (2000 iters) | `2000` | `IR interpreter: var not found user/deftype*` | `2000` |
| `deftype` in a hot fn | `2000` | same error | `2000` |
| `cljrs compile` a `deftype` program | `9` | `nil`, exit 0 | `9` |
| `cljrs compile` a `reify` program | `7` | `nil`, exit 0 | `7` |

Also confirmed fixed: the `builtin_stub_nil` entries are gone and `deftype*`/`protocol*` have their own; the three crate READMEs are updated; and `(deftype Bad)` now reads `expected 2+, got 1`. The `IMPLICIT_MACRO_ARGS` fix in `select_arity` is the right shape — it repairs every macro's arity message, not just this family's.

All 70 tests in the seven affected targets pass locally (`defmethod_cross_ns`, `defmulti_macro`, `defprotocol_macro`, `extend_macros`, `dispatch_family_expansion`, `datatype_in_fn_tiered`, `deftype_types`). Spot-checked by hand and correct: `extend-type` multi-arity collapsing, a variadic protocol method, `:extend-via-metadata` dispatch resolving through the protocol's own namespace, `defmulti`'s docstring/attr-map/`:default nil` head shapes.

---

### New finding — runtime startup is 2.5× slower

Every runtime construction pays for the larger `bootstrap.cljrs`. Measured on an empty program, release builds, interleaved runs, 8 reps each:

```
PR    median 47–50 ms   (min 45)
main  median 17–24 ms   (min 16)
```

Peak RSS is unchanged (25.3 MB both), so this is time, not memory. It lands on every `cljrs run`/`eval`/`repl` invocation, on every AOT binary's startup, and on every `Runtime` the test suite builds — which is per proptest *case* in the property-test targets, so it multiplies there.

`bootstrap.cljrs` grew 15% (1616 → 1868 lines) for a 170% startup increase, so it is worth profiling rather than assuming it is proportional — the hoisted forms now macroexpand through Clojure where they used to be Rust special forms, including bootstrap's own uses of them.

Relevant context, though I can't pin it on the PR: `Build, test, lint` passed in 25m05s on the previous head (`db36a1d`) and this head's job was killed at 26m15s by a runner shutdown while `prop_splice_agrees_across_eval_quote_and_syntax_quote` and `prop_splice_evals_as_inline_in_every_container` were mid-run. A runner shutdown is an infra event, so that alone proves nothing — but the job is now at the edge of its budget and this regression pushes the same direction. Worth a re-run, and if it dies again the startup cost is the first thing I'd look at. (That proptest takes 219s by itself in a debug build here, and two test binaries were SIGKILLed locally under memory pressure in this container, which I do **not** attribute to the PR.)

### Pre-existing bug this PR is well placed to fix

Not a regression — I measured it on `main` too — but `defmethod` is being rewritten here, so it is a two-word fix now and a bespoke one later.

The expansion names the impl fn after the multimethod:

```clojure
(list 'add-method mname dispatch-val
      (cons 'fn (cons (symbol (name mname)) (cons params body))))
```

`apply_fn` self-binds any named fn (`crates/cljrs-runtime/src/interp/apply.rs:594`), so inside a method body the multimethod's own name resolves to *that impl fn* rather than re-dispatching:

```clojure
(defmulti area :kind)
(defmethod area :square [s] (* (:side s) (:side s)))
(defmethod area :composite [s] (reduce + (map area (:parts s))))
(area {:kind :composite :parts [{:kind :square :side 2} {:kind :square :side 3}]})
```

`13` in Clojure; `0` on both `main` and this branch — `area` in the `:composite` body is the `:composite` impl, so each part re-enters it with no `:parts` and reduces to zero. A delegating method is worse: `(defmethod kindof :a [s] (str "a then " (kindof (assoc s :k :zzz))))` recurses into itself ~150 deep instead of dispatching once.

`main` reached the same place from the other side — `eval_defmethod` passed `fn_name = Some(multi_name)` to `CljxFn::new`. Clojure's own `defmethod` emits an anonymous `fn` for exactly this reason, so `(cons 'fn (cons params body))` fixes it; a gensym-suffixed name would also work if the name matters for stack traces, since a user can't collide with it.

### Non-blocking

**Five implementation helpers are public in `clojure.core`.** `defmethod-pin`, `protocol-method-spec`, `protocol-body`, `extend-spec-groups` and `extend-impl-map` all answer `resolve` from user code. `defn-` works in this runtime (checked), and these are internals of the hoist rather than API.

**`extend` doesn't validate method names.** `(extend 'T P {:typo f})` registers an entry no dispatch will ever read, silently. `eval_extend_type` didn't check either, so nothing is lost — but `extend` is new public API with `protocol-fn` right next to it already resolving names against the protocol's spec, so the check is cheap here.

**`extending_something_undefined_says_that_instead`** now asserts `UnboundSymbol`, which is a reasonable call and well argued in the comment — but the test name still promises the old wording. Worth renaming to match.

**`extend`'s odd-pair error** reports `got: args.len()`, which counts the type tag along with the pairs.

---

PR #384, opened 2026-09-09
AOT harness builds share one cargo target dir

BuddhiLW:
Every AOT link generates a harness crate and runs `cargo build` in it with no `CARGO_TARGET_DIR`, so each harness gets an empty `target/` and recompiles the whole cljrs dependency tree. Nothing is reused between two links.

That is why `cljrs compile` costs a from-scratch build every time, and why the AOT test suite spends nearly all of its wall time compiling the same dependencies over and over.

## The change

Point the harness build at one directory: `CLJRS_AOT_TARGET_DIR` if set, else `~/.cljrs/cache/aot-target`. With no `HOME` and no override the variable is left unset, which is exactly the current behaviour.

## Measured

Ungated `aot_e2e`, the same six tests:

```
before                            474.6s
after, cold (cljrs-compiler changed)  91.2s   one full build, five reuses
after, warm                          3.5s
```

A `cljrs compile` of a second file drops from 93.8s to 4.7s. On disk it is one 427M cache instead of a full dependency tree per harness build.

This should also make the gated `aot_full_test` suite tractable locally — its cost is roughly one dependency-tree compile per test today.

## Notes

Cargo locks a target directory, so concurrent harness builds queue rather than corrupt one another; `aot_e2e` already serializes itself behind `AOT_LOCK` in any case.

`harness_target_dir_from` takes the two values as arguments instead of reading the environment, so the resolution rule is testable without `set_var` — which is UB once libtest has started a second thread. Three unit tests cover it: two harnesses resolving to the same path, the explicit override winning, and the nothing-to-resolve case leaving `CARGO_TARGET_DIR` unset. Mutation-checked — removing the override fails the override test and nothing else.

`cargo test -p cljrs-compiler` is green (105 lib + all integration binaries, `test_harness_e2e` included), `cargo fmt --all --check` clean, `cargo clippy -p cljrs-compiler --all-targets` clean.

### DIFF
--- a/crates/cljrs-compiler/src/aot.rs
+++ b/crates/cljrs-compiler/src/aot.rs
@@ -2448,10 +2448,43 @@ fn link_with_cargo_test_harness(
     let built = harness_bin_from_cargo_stdout(&stdout)?;
     std::fs::copy(&built, out_path)?;
 
-    // Keep the harness directory for debugging
+    // Kept on purpose: `tests/test_harness_e2e.rs` inspects the generated
+    // project.  The ordinary compile path (`link_with_cargo`) discards its
+    // harness instead.
     eprintln!("[aot] harness directory kept at {}", harness_dir.display());
 
     Ok(())
+}
+
+/// Where harness builds put their intermediate artifacts.
+///
+/// One directory for every harness, so the cljrs dependency tree is compiled
+/// once rather than once per link. `CLJRS_AOT_TARGET_DIR` overrides it;
+/// otherwise `~/.cljrs/cache/aot-target`, or the harness's own `target/` when
+/// there is no home directory to cache under.
+///
+/// Cargo locks a target directory, so concurrent harness builds queue rather
+/// than corrupt one another.
+fn harness_target_dir() -> Option<PathBuf> {
+    harness_target_dir_from(
+        std::env::var_os("CLJRS_AOT_TARGET_DIR"),
+        std::env::var_os("HOME"),
+    )
+}
+
+/// [`harness_target_dir`] as a function of its inputs, so the rule can be
+/// exercised without mutating process environment (`set_var` is UB once a
+/// second thread exists, and libtest runs tests on many).
+fn harness_target_dir_from(
+    explicit: Option<std::ffi::OsString>,
+    home: Option<std::ffi::OsString>,
+) -> Option<PathBuf> {
+    if let Some(explicit) = explicit.filter(|v| !v.is_empty()) {
+        return Some(PathBuf::from(explicit));
+    }
+    home.filter(|v| !v.is_empty())
+        .map(PathBuf::from)
+        .map(|home| home.join(".cljrs").join("cache").join("aot-target"))
 }
 
 /// Run `cargo build --release --message-format=json` in the harness directory.
@@ -2465,6 +2498,9 @@ fn cargo_build_harness_release(
         .arg("--message-format=json");
     if offline {
         cmd.arg("--offline");
+    }
+    if let Some(target_dir) = harness_target_dir() {
+        cmd.env("CARGO_TARGET_DIR", target_dir);
     }
     let output = cmd.current_dir(harness_dir).output()?;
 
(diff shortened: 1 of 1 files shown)


crates/cljrs-compiler/src/aot.rs
```diff
@@ -2463,6 +2496,9 @@ fn cargo_build_harness_release(
     if offline {
         cmd.arg("--offline");
     }
+    if let Some(target_dir) = harness_target_dir() {
+        cmd.env("CARGO_TARGET_DIR", target_dir);
```
csm, line 2500: Every harness is generated as the same package — `name = "cljrs-aot-harness"`, `version = "0.1.0"`, same dependency set (`build_harness`, ~line 2071). Before this PR each harness had its own `target/`, so that didn't matter. Sharing one target dir means harnesses at *different paths* land in the same Cargo fingerprint slot and uplift to the same `release/cljrs-aot-harness`.

The shared cache after my runs holds two `cljrs_aot_harness-<hash>` artifacts for ~7 distinct harness crates:

```
$ ls aot-cache/release/deps | grep cljrs_aot_harness
cljrs_aot_harness-bcaf008b70333cb8
cljrs_aot_harness-ff75294bd52e11f6
```

Minimal repro of the consequence — two crates, same name/version, different directories, one `CARGO_TARGET_DIR`:

```
$ cd h1 && CARGO_TARGET_DIR=$S/shared cargo build --release
$ $S/shared/release/cljrs-aot-harness
h1
$ cd ../h2 && CARGO_TARGET_DIR=$S/shared cargo build --release --message-format=json | grep -o '"fresh":[a-z]*'
"fresh":true
$ $S/shared/release/cljrs-aot-harness
h1            # ← h2's build reported this path as its `executable`; the bytes are h1's
```

`harness_bin_from_cargo_stdout` returns that path, so `std::fs::copy` would write the wrong binary to `out_path`. Touch `h2/src/main.rs` so it is newer than the artifact and the same build goes `"fresh":false` and links correctly — freshness here is decided by mtime comparisons against the *other* harness's recorded dep-info.

I could not make this fire through the CLI. Both paths stay correct because the generated `build.rs` embeds the harness's absolute `.o` path and both files are written immediately before the build, so they are always newer than any prior artifact:

```
compile a.cljrs -o a/abin   → I AM A
compile b.cljrs -o b/bbin   → I AM B   (1.8s, reused deps)
compile a.cljrs -o a/bin3   → I AM A2  (alternating a/b/a stays correct)
compile --test ta -o ta/bin → marker-a only, with tb's harness dir still on disk
compile --test tb -o tb/bin → marker-b only
```

So this isn't a bug today — it's an invariant the PR newly depends on ("regenerated harness files are always newer than the cached artifact") that nothing states or tests. It breaks under clock skew, which is exactly what `CLJRS_AOT_TARGET_DIR` invites: point several machines or containers at one shared cache and a harness written by a slow-clocked host looks stale against another host's artifact, and you silently get the other program's binary. Cargo's target-dir lock doesn't help — it serializes builds, not identities. The same collision makes `release/cljrs-aot-harness` a shared uplift destination, so two concurrent `cljrs compile` runs also race between cargo exiting and `std::fs::copy` (narrow, but it was impossible before).

Giving each harness a distinct package name (e.g. `cljrs-aot-harness-<hash of harness_dir>`, with `AOT_HARNESS_BIN` matching accordingly) gives every harness its own fingerprint slot and uplift path and keeps the whole win — the dependency units are keyed by their own package ids and stay shared; only the ~1s harness crate itself stops being reused.

---

crates/cljrs-compiler/src/aot.rs
```diff
@@ -2445,12 +2445,45 @@ fn link_with_cargo_test_harness(
...
+/// Where harness builds put their intermediate artifacts.
+///
+/// One directory for every harness, so the cljrs dependency tree is compiled
+/// once rather than once per link. `CLJRS_AOT_TARGET_DIR` overrides it;
+/// otherwise `~/.cljrs/cache/aot-target`, or the harness's own `target/` when
+/// there is no home directory to cache under.
+///
+/// Cargo locks a target directory, so concurrent harness builds queue rather
+/// than corrupt one another.
+fn harness_target_dir() -> Option<PathBuf> {
```
csm, line 2465: `harness_target_dir` doesn't consult an inherited `CARGO_TARGET_DIR`, and `cargo_build_harness_release` sets it unconditionally — so whenever `HOME` is set (i.e. nearly always), an exported `CARGO_TARGET_DIR` is silently overridden. Before this PR the harness build inherited it.

That also makes the doc comment above inaccurate in one case: with no `HOME` and an inherited `CARGO_TARGET_DIR`, the fallback is that inherited directory, not "the harness's own `target/`".

Resolution order `CLJRS_AOT_TARGET_DIR` → inherited `CARGO_TARGET_DIR` → `~/.cljrs/cache/aot-target` costs one more `var_os` and keeps a caller's explicit choice working. If ignoring it is deliberate, worth saying so in the doc comment.

---

csm (commented):
Reviewed the diff (one file, +79/-1) and reproduced the change end to end on `c43837b`.

## The win is real

Ran `cljrs compile` against a scratch `CLJRS_AOT_TARGET_DIR`:

| | |
|---|---|
| first file, cold cache | 2m59s |
| second file, different directory | 1.8s |
| test harness, cold | 50s |
| test harness, different directory | 1.5s |

Same shape as the numbers in the description. `cargo test -p cljrs-compiler --lib` passes including the three new unit tests; `cargo fmt --all --check` is clean.

The shape of the change is right: both link paths funnel through `cargo_build_harness_release`, so the ordinary compile and the test harness both pick it up; taking the two values as arguments to keep `set_var` out of the tests is the correct call.

## Two things to look at

1. **Every harness is the same Cargo package, so one target dir means one fingerprint slot.** Not a bug today — I could not make it fire through the CLI — but the correctness now rests on an mtime invariant that nothing states or tests, and it breaks under clock skew across a shared cache, which is the case `CLJRS_AOT_TARGET_DIR` exists to enable. Detail, repro and a suggested fix inline.
2. **An inherited `CARGO_TARGET_DIR` is silently overridden.** Inline.

## Smaller

- **The new env var and the new cache directory aren't documented anywhere.** `crates/cljrs-compiler/README.md` has a "Harness dependency resolution" section that documents `CLJRS_WORKSPACE_ROOT` in exactly this spirit, and the root `README.md` has an env-var table. CLAUDE.md asks for READMEs to move in the same commit.
- **Cache growth is unbounded.** My cache reached 992M over a handful of links, and nothing prunes it or exposes a way to clear it. `~/.cljrs/cache/{git,dylibs}` set the same precedent so this is consistent, but this one is much larger and worth a sentence in the README next to the env var.
- **Nit:** this is the third hand-rolled `HOME` → `~/.cljrs/cache/...` resolution (`cljrs/src/native/pinned.rs:822`, `cljrs-project/src/vcs.rs:113`). Not this PR's job to unify, but a shared helper is getting overdue. Note the other two fall back to `"."` where this one falls back to `None` — the choice here is the better one.

## Unrelated, found while testing

Re-running `cljrs compile --test <dir>` in a directory that still holds a `.cljrs-aot-test-harness/` from a previous run fails: test discovery walks into the kept harness and tries to resolve its bundled sources.

```
[aot] discovered 2 test namespace(s)
Error: eval/lowering error: Could not find source for namespace
       .cljrs-aot-test-harness.src.bundled-0
```

Pre-existing on `main`, nothing to do with this change — flagging it since the diff touches the comment explaining why that directory is kept.

---

PR #386, opened 2026-09-09
recur across an await no longer deadlocks

BuddhiLW:
## The bug

```clojure
;; works — two sequential takes
(let [x (<? c) y (<? c)] [x y])          ;=> [1 2]

;; hangs forever — the identical takes, reached by recur
(loop [acc [] i 0]
  (if (>= i 2) acc (recur (conj acc (<? c)) (inc i))))
```

No output, no error, no exit: killed only by timeout. An `^:async` fn calling itself worked, and the same takes outside a loop worked, which made the failure look arbitrary rather than structural.

## Cause

`eval_async` owns `await`, `do`, `if`, `let*`, `loop*` and `try`, and hands every **other** special form to the synchronous evaluator:

```rust
other if is_special_form(other) => return eval(&expanded, env),
```

`recur` is a special form, so it took that branch — and with it, so did its *arguments*. Sync `eval_await` on a future that is still `FutureState::Running` does:

```rust
FutureState::Running => { guard = f.get().cond.wait(guard).unwrap(); }
```

That blocks the thread. The executor is a `current_thread` runtime on that same thread, so the future it is waiting for can never be driven. The loop deadlocks on the value it asked for.

`recur` reads like a form the sync evaluator could own, because it yields no value of its own — it signals `EvalError::Recur` to a trampoline. But its arguments are ordinary expressions, and they are exactly where the await lives.

## Fix

`eval_recur_async` evaluates the arguments with `eval_async`, roots the accumulated values across each yield, and signals the trampoline exactly as before. Both recur targets are covered: the `loop*` header (`eval_loop_async`) and the enclosing `^:async` fn (`run_async_fn`).

## Why this shape matters

Take a chunk, append it, test, go again — that is every stream client. Reading a newline-framed reply, an HTTP body to `Content-Length`, or a chunked transfer all want this loop, and each one deadlocked when written the obvious way. The workaround was a self-recursive `^:async` fn, which is not what a Clojure programmer writes and which nothing suggested.

## Testing note

The regression cannot be guarded with `tokio::time::timeout`. The hang is a blocking `Condvar` wait on the executor's own thread, so the timer future is never polled again and the timeout cannot fire — wrapping it would simply hang too. Each case therefore runs on a spawned thread with a `recv_timeout` watchdog on the test thread.

Before the fix, both awaiting cases trip the watchdog at 20s. After, the file runs in 0.03s.

https://claude.ai/code/session_01XvnAiUQ3wkvz5zAjBHWkt5


crates/cljrs-async/README.md
```diff
@@ -162,11 +163,14 @@ user-level catch. Every `spawn_future` captures the complete active meter stack
...
-but is fragile when the awaited value is not yet available. Yielding `try`/`catch` in
-`eval_async` is the remaining follow-up.
+**Which forms yield:** `eval_async` has a yielding arm for `await`, `do`, `if`, `let`/`let*`,
+`loop`/`loop*`, `recur`, `try` (body, `catch` handlers and `finally`), collection literals, and
+function-call arguments. Every *other* special form is delegated to the synchronous evaluator,
+which evaluates its sub-expressions on the blocking `await` path. On the single-threaded
+`LocalSet` that deadlocks as soon as the awaited future is not already settled, because the task
+that would settle it cannot run while the executor thread is parked. So a special form that can
+carry an `await` anywhere inside it needs its own arm here; a fallthrough is a latent hang, not a
+slow path. `binding`, `and`/`or` and `throw` are the remaining delegated forms.
```
csm, line 173: This list is incomplete, and this paragraph is the one place a reader will look to find out what is still dangerous.

`SPECIAL_FORMS` (`crates/cljrs-runtime/src/builtins/special.rs:2`) has 39 entries. After this PR, `eval_async` owns `await`, `do`, `if`, `let`/`let*`, `loop`/`loop*`, `recur`, `try`. Everything else takes the `is_special_form` fallthrough — not just `binding`, `and`/`or`, `throw`. The ones that evaluate a sub-expression, and so can carry an `await`:

`def`, `defonce`, `set!`, `throw`, `and`, `or`, `.`, `letfn`, `binding`, `with-out-str`, plus the body-evaluating `defmethod`/`reify`/`extend-type`/`extend-protocol` forms.

`def` is the notable omission — `(def x (await …))` is a shape someone will write at an async top level on day one. I confirmed three of these hang on this branch, with a concurrent producer so the take genuinely pends:

```clojure
(def ch (chan 1))
(go (await (put! ch 42)))
(def x (await (take! ch)))        ; watchdog at 6s
(and true (await (take! ch)))     ; watchdog at 6s
(throw (ex-info "x" {:v (await (take! ch))}))  ; watchdog at 6s
```

Same mechanism as the `recur` bug this PR fixes. Suggest just naming the full set here, since the paragraph's own rule — "a fallthrough is a latent hang, not a slow path" — makes the list load-bearing rather than decorative.

---

crates/cljrs-async/src/eval_async.rs
```diff
@@ -462,6 +470,23 @@ async fn eval_loop_async(args: &[Form], env: &mut Env) -> EvalResult {
...
+/// catches. Mirrors the synchronous `eval_recur` (`cljrs_runtime::interp::special`)
+/// apart from the evaluator used for the arguments.
+async fn eval_recur_async(args: &[Form], env: &mut Env) -> EvalResult {
+    let mut vals: Vec<Value> = Vec::with_capacity(args.len());
+    for form in args {
+        // Root the arguments already evaluated: each remaining `await` is a
+        // yield point at which a GC cycle may run.
+        let _vals_root = cljrs_runtime::env::gc_roots::root_values(&vals);
+        let val = Box::pin(eval_async(form, env)).await?;
+        vals.push(val);
```
csm, line 485: Non-blocking note on the rooting, since it's subtler than it looks.

`root_values` stores `(vals.as_ptr(), vals.len())` — a raw pointer, not a borrow (`gc_roots.rs:112`). `_vals_root` is still live when `vals.push(val)` runs on the line below, so the entry would dangle if that push reallocated. It can't, because `with_capacity(args.len())` is exact and the loop pushes at most `args.len()` times — the safety here rests entirely on that. `VALUE_ROOTS` is thread-local and scanned only by its own thread, so there's no cross-thread window either.

This is the same shape as `eval_call_async` (lines 511-515, 532-535), so it's the house idiom and I wouldn't change it. Worth a half-line on the `with_capacity` if you're touching the comment anyway.

---

csm (commented):
Reviewed at 44001e1. The fix is correct and the diagnosis in the description matches what the code does.

**What I verified locally**

- `eval_recur_async` mirrors the sync dispatch exactly: `special.rs:34` calls `eval_recur(args, env)` with `args = forms[1..]`, and the new arm passes the same slice. Only the evaluator for the arguments changes; the `Err(EvalError::Recur(vals))` signal is byte-identical, so both trampolines (`eval_loop_async:466`, `run_async_fn:149`) keep working unchanged. `(recur)` with zero args still yields `Recur(vec[])`.
- The `"recur"` arm sits above the `other if is_special_form(other)` arm, so it actually takes precedence.
- The 4 new tests all pass (0.04s). I then removed just the dispatch line and re-ran: all 4 fail on the 20s watchdog. They are genuine regression tests, not tests that pass either way — including `async_recur_still_enforces_loop_arity`, which I half expected to be fix-independent.
- Full `cljrs-async` suite green (18/42/4/1/6/11/2/6/7/6/8 across targets), `cargo clippy -p cljrs-async --all-targets` clean, `cargo fmt --check` clean.

The testing note in the description is right and worth keeping: `tokio::time::timeout` genuinely cannot catch this, because the blocking `Condvar` wait parks the thread that would poll the timer. The spawned-thread + `recv_timeout` watchdog is the only shape that reports the hang instead of joining it, and the detached-on-purpose comment explains why the threads aren't joined.

**One thing I'd change**

The README's "remaining delegated forms" sentence is wrong — inline comment on `README.md:173`. `def`, `set!`, `.`, `letfn`, `with-out-str` and the type/protocol forms are also delegated, and I confirmed `(def x (await …))` deadlocks on this branch the same way `recur` did. Docs-only, doesn't affect the fix, but this PR is what promotes that list from a footnote to the reader's map of where the hangs are.

Plus a non-blocking note on the rooting invariant in `eval_recur_async`.

---

PR #390, opened 2026-09-12
Name each extension's init symbol after its crate

BuddhiLW:
An extension crate's C-ABI entry point is `#[no_mangle]`, and both in-tree
plugins spelled it `cljrs_init`. An unmangled symbol is global to the whole
process image, so those are the *same* symbol.

The interesting part is what that does, because it is not what the name
"duplicate symbol" suggests. Linking both crates into one binary does **not**
fail:

```console
$ nm target/debug/deps/<test-binary> | grep cljrs_init
00000000002ef7b0 T cljrs_init          # one definition, not two
```

The linker resolves every reference to a single definition, so
`cljrs_base64::cljrs_init` and `cljrs_blake3::cljrs_init` evaluate to the same
address. One plugin's registration silently replaces the other's and the second
namespace is simply never defined. No error at build time, none at run time,
only a namespace that is missing.

Measured before this change, with both crates linked into one test binary:

```
assertion `left != right` failed
  left: 95096257017776
 right: 95096257017776
```

### Not yet reachable, which is why it is worth fixing now

`cljrs.edn` carries a single `:rust :init`, so today an AOT build links at most
one plugin and nothing collides. It becomes reachable the moment `:rust :init`
accepts more than one crate. Until then the bug is invisible, and when it
arrives it presents as a missing namespace rather than as a link error, which is
a bad thing to debug for the first time under deadline.

### The fix

Name the export after its crate: `cljrs_init_cljrs_base64`,
`cljrs_init_cljrs_blake3`, with the convention documented as
`cljrs_init_<crate>`. Uniqueness then follows from crate names already being
unique, rather than from everyone remembering a rule.

`tests/extension_init_symbols.rs` pins it. It is the only test binary that links
two plugin crates, which is why `cljrs-blake3` is now a dev-dependency of
`cljrs-compiler`:

```console
$ nm target/debug/deps/extension_init_symbols-<hash> | grep cljrs_init
0000000000287bf0 T cljrs_init_cljrs_base64
000000000025a7b0 T cljrs_init_cljrs_blake3
```

### Backwards compatible

Nothing hardcodes the symbol name. `native/mod.rs` takes the last `::` segment
of `:rust :init` and looks *that* up, so the name is data: an existing extension
keeps its own `cljrs_init` and keeps working. Only the convention taught to new
extensions changes.

Docs moved with it: `project-setup.md` gains a "Name the symbol after your
crate" section explaining the shadowing mechanism, and the examples in 11 book
pages, 4 crate READMEs, `VERSIONING.md` and the `:rust :init` doc comments
follow. Both plugins' own `cljrs.edn` are updated, without which the rename
would have broken dylib loading.

`cargo fmt --check`, `cargo clippy --workspace -- -D warnings` and
`cargo test --workspace` (165 suites, 1507 tests) all clean.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -825,6 +825,7 @@ version = "0.1.0"
 dependencies = [
  "cljrs-async",
  "cljrs-base64",
+ "cljrs-blake3",
  "cljrs-charset",
  "cljrs-gc",
  "cljrs-io",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -61,6 +61,7 @@ cljrs-charset      = { path = "crates/cljrs-charset",      version = "0.1.0", de
 cljrs-lsp          = { path = "crates/cljrs-lsp",          version = "0.1.0" }
 cljrs-nrepl        = { path = "crates/cljrs-nrepl",        version = "0.1.0", default-features = false }
 cljrs-base64       = { path = "crates/cljrs-base64",       version = "0.1.0", default-features = false }
+cljrs-blake3       = { path = "crates/cljrs-blake3",       version = "0.1.0", default-features = false }
 cljrs-tx           = { path = "crates/cljrs-tx",           version = "0.1.0", default-features = false }
 
 tracing            = "0.1.44"
--- a/VERSIONING.md
+++ b/VERSIONING.md
@@ -138,7 +138,7 @@ provenance** (the commit it was built from, registered via
 {:deps
  {my.native.lib {:git/url   "https://github.com/user/my-native-lib"
                  :git/sha   "abc1234ef"
-                 :rust/init "my_native_lib::cljrs_init"
+                 :rust/init "my_native_lib::cljrs_init_my_native_lib"
                  :rust/load :dylib}}}
 ```
 
--- a/crates/cljrs-base64/README.md
+++ b/crates/cljrs-base64/README.md
@@ -8,13 +8,13 @@ Exposes standard and URL-safe Base64 encode/decode functions to Clojure code und
 
 ## Status
 
-Phase 1 — implemented. Linked statically into the `cljrs` binary via the `base64` feature (enabled by default) and also loadable as a dynamic plugin via the `cljrs_init` FFI entry point.
+Phase 1 — implemented. Linked statically into the `cljrs` binary via the `base64` feature (enabled by default) and also loadable as a dynamic plugin via the `cljrs_init_cljrs_base64` FFI entry point.
 
 ## File layout
 
 | File | Description |
 |---|---|
-| `src/lib.rs` | All implementation: byte-conversion helpers, `init`, `register`, and the `cljrs_init` FFI entry point |
+| `src/lib.rs` | All implementation: byte-conversion helpers, `init`, `register`, and the `cljrs_init_cljrs_base64` FFI entry point |
 
 ## Public API
 
@@ -30,7 +30,7 @@ pub fn register(registry: &mut Registry);
 
 /// C-ABI entry point for dynamic plugin loading.
 #[no_mangle]
-pub unsafe extern "C" fn cljrs_init(registry: *mut Registry);
+pub unsafe extern "C" fn cljrs_init_cljrs_base64(registry: *mut Registry);
 ```
 
 ### Clojure functions
--- a/crates/cljrs-base64/cljrs.edn
+++ b/crates/cljrs-base64/cljrs.edn
@@ -3,4 +3,4 @@
  :paths ["src"]
  :test-paths ["test"]
  :rust {:crate "."
-        :init "cljrs_base64::cljrs_init"}}
+        :init "cljrs_base64::cljrs_init_cljrs_base64"}}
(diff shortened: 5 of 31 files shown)


csm (commented):
## Verdict

The rename is correct and worth merging. The stated **reasoning is wrong in two ways, both of which make the bug worse than this PR claims**, and the wrong explanation is now written verbatim into four user-facing places. Worth fixing the prose before merge.

## 1. The bug is reachable today, not hypothetical

The description says: *"`cljrs.edn` carries a single `:rust :init`, so today an AOT build links at most one plugin and nothing collides."*

That is only true of *user-declared* plugins. The AOT harness also links the CLI's default extension set:

- `crates/cljrs/Cargo.toml:10` — `default = ["async", "net", "charset", "base64", ...]`
- `crates/cljrs/src/extensions.rs:64-69` — pushes `cljrs-base64` into `default_set()`
- `crates/cljrs-compiler/src/aot.rs:2062-2068` — every name from `extensions.crate_names()` lands in the generated harness's `[dependencies]`, alongside the user crate added at `:2053-2057`

So **every** `cljrs compile` links `cljrs-base64`, which before this PR exported `#[no_mangle] cljrs_init`. A user project following the previously documented `:init "my_project::cljrs_init"` collides with it on the next build — and that spelling is what every doc page in this repo recommended until this PR.

No AOT test exercises a user crate with `:rust :init` (`grep -rn RustConfig crates/*/tests/` is empty), which is why CI never hit it.

## 2. "Does not fail to link" is wrong in the case that actually occurs

I reproduced both behaviors with minimal crates. Which one you get depends on whether the colliding archive member is dragged in for an unrelated reason.

**Only `cljrs_init` referenced** → silent merge, as described here:

```
a=562118302b40 b=562118302b40 same=true
0000000000014b40 T cljrs_init      # one definition
```

The linker resolves the single undefined `cljrs_init` from the first archive that defines it and never pulls the second member.

**The harness's actual shape** — crate A also exports a normal `init()` that the harness calls, which is exactly `cljrs_base64::init`, so A's object is pulled in regardless:

```
rust-lld: error: duplicate symbol: cljrs_init
>>> defined at b/src/lib.rs:4 ... in archive libsymb.rlib
>>> defined at a/src/lib.rs:5 ... in archive libsyma.rlib
```

A hard link failure. Whether `init` and `cljrs_init` land in the same codegen unit decides which outcome a given build gets, so it is also non-deterministic across opt levels and CGU partitioning.

The claim as written ("No error at build time, none at run time, only a namespace that is missing") now appears in `crates/cljrs-base64/src/lib.rs:134-145`, `crates/cljrs-compiler/README.md:415-420`, `docs/book/src/rust-interop/project-setup.md:88-99` and `crates/cljrs-compiler/tests/extension_init_symbols.rs:6-13` — four copies of an explanation that will mislead the next person debugging a `duplicate symbol` error, which is the one message they are most likely to actually see.

## 3. Test coverage misses the pair that collides

`extension_init_symbols.rs` pins base64 vs. blake3. Those two never share a binary in the field. The pair that does is `cljrs-base64` (always linked) vs. an arbitrary *user* crate, which a unit test cannot name. The test is still worth keeping — it catches symbol merging regardless of spelling, including an `#[export_name]` reintroduction — but it does not cover the live case. An AOT e2e fixture with a `:rust :init` crate would.

## 4. Clippy regression

`cargo clippy -p cljrs-compiler --all-targets -- -D warnings` fails on the new test:

```
error: direct cast of function item into an integer   (x2)
error: could not compile `cljrs-compiler` (test "extension_init_symbols")
```

The workspace under `--all-targets` has exactly one other warning, so this is 2 of 3. CI runs `cargo clippy --workspace -- -D warnings` without `--all-targets` (`.github/workflows/ci.yml:35`), so it passes — the "clippy clean" claim holds only on that technicality. Fix is `as *const () as usize` on `extension_init_symbols.rs:15-16`.

## 5. Two doc examples left behind

`docs/book/src/cli/deps.md:168` and `:181` still read `:rust/init "my_lib::cljrs_init"`, while the equivalent `:rust/init` examples in `VERSIONING.md:141`, `docs/book/src/language/versioned-symbols.md:92` and `crates/cljrs-project/README.md:230` were converted.

Related: for `:rust/load :dylib` deps the init fn is never dlsym'd — `crates/cljrs/src/native/pinned.rs:702` generates a wrapper that calls `{pkg}::{init_tail}(registry)` as a plain Rust call and exports `cljrs_dylib_init` itself. `#[no_mangle]` is not needed there and no collision is possible, so all the `:rust/init` renames are cosmetic consistency. Fine either way, but it means deps.md is the only inconsistency worth closing.

## Verified locally

`cargo fmt --check` clean. `cargo test -p cljrs-compiler --test extension_init_symbols` passes. The backwards-compatibility claim holds: `crates/cljrs/src/native/mod.rs:91` takes the last `::` segment, `pinned.rs:265-270` splits on the first — both treat the name as data, and the `cljrs_init` fixtures in `pinned_dylib_e2e.rs` and `config/parse.rs` were correctly left alone.

---

